PR #11, opened 2025-12-18
Add testID support to shared input components (E2E readiness)

valerymaksimovich1986:
This PR adds testID support to shared input components to improve mobile E2E automation.
IDs are now available both on the wrapper and on the underlying TextInput.

No functional or UI changes were introduced.

---
  - https://app.asana.com/0/0/1212508942478666

### DIFF
--- a/e2e-mobile-tests/tests/locators/SignUpLocators.ts
+++ b/e2e-mobile-tests/tests/locators/SignUpLocators.ts
@@ -1,5 +1,185 @@
 const signUpLocators = {
+  /* =====================================================
+        CREATE MASTER PASSWORD
+  ===================================================== */
 
+  createPasswordScreenAndroid: '~create_password_screen',
+  createPasswordScreeniOS: '~create_password_screen',
+
+  createPasswordLogoAndroid: '~create_password_logo',
+  createPasswordLogoiOS: '~create_password_logo',
+
+  createPasswordTitleAndroid: '~create_password_title',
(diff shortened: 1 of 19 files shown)


src/libComponents/PasswordField/index.test.js
```diff
@@ -171,4 +171,28 @@ describe('PasswordField Component', () => {
 
     expect(input.props.secureTextEntry).toBe(true)
   })
+
+  test('passes inputTestID to password input', () => {
+    const { getByTestId } = render(
+      <ThemeProvider>
```
eugeneglova: I've noticed every test includes theme provider, maybe it's good idea to create some helper like `renderTheme()` and use it instead of `render(<PasswordField {...props}>)`

e2e-mobile-tests/tests/locators/SignUpLocators.ts
```diff
@@ -1,5 +1,185 @@
 const signUpLocators = {
+  /* =====================================================
+        CREATE MASTER PASSWORD
+  ===================================================== */
 
+  createPasswordScreenAndroid: '~create_password_screen',
+  createPasswordScreeniOS: '~create_password_screen',
```
eugeneglova: I am not an expert in e2e mobile tests, but seems like a lot of duplication here, is there any optimizations to do not add iOS and Android ? if it's the same value and only use it if it's different

PR #53, opened 2026-01-02
Enhance version comparison

torkinos:
Enhance version comparison to handle non-numeric characters in current version. e.g. 1.1.1-dev

### DIFF
--- a/src/hooks/useVersionCheck.js
+++ b/src/hooks/useVersionCheck.js
@@ -13,7 +13,10 @@ import { logger } from '../utils/logger'
  * @returns {boolean} True if latest > current
  */
 export const compareVersions = (current, latest) => {
-  const currentParts = current.split('.').map(Number)
+  const currentParts = current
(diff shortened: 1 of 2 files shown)


src/hooks/useVersionCheck.js
```diff
@@ -13,7 +13,10 @@ import { logger } from '../utils/logger'
  * @returns {boolean} True if latest > current
  */
 export const compareVersions = (current, latest) => {
-  const currentParts = current.split('.').map(Number)
+  const currentParts = current
```
alvaropaco: Here I have some defnsive (Safety) recommendations:

- The function assumes current and latest are always valid strings.
- Passing null, undefined, or non-string values will throw at runtime.
- Number() silently converts invalid segments to NaN, which leads to undefined behavior in comparisons.

alvaropaco (changes requested):

PR #117, opened 2026-02-02
Sync dev to main

torkinos:


.github/workflows/build-and-publish.yml
```diff
@@ -70,8 +69,11 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.PAT_TOKEN }}
 
-      - name: Generate worklet bundle for Android
-        run: npm run bundle:android
+      - name: Generate worklet bundle and translations
+        run: npm run build
+
+      - name: Generate worklet bundle for Autofill
+        run: npm run bundle:autofill
```
ndrkltsk: duplicated step

ndrkltsk: actually, we can remove `npm run bundle:autofill` entirely, as its executed by `npm run build` which executes bundle-bare

PR #119, opened 2026-02-03
CopyButton attached to fields

harrisaar-noxtton:
# Changes

1. Implemented the copy feature for the note form
2. If "is copy enabled" is set to false, then copy icons will not be rendered.

# Code

Implemented `CopyButton` component similar to the Desktop and Extension project which handles the rendering and copying logic.


torkinos (changes requested):
@harrisaar-noxtton please fix failing tests

torkinos (approved):

PR #284, opened 2026-04-02
New Vault create Redesign V2 design

alvaropaco:
### Requirements
- Implement the revamp for the Share Personal Vault screen based on the provided Figma design.
- Ensure the Share Vault capability is accessible from:
  Home > Selecting Vault -> "X" Vault Actions > Share Access
- Keep the implementation aligned with the existing mobile vault invite flow and current SDK capabilities.
- Maintain consistency with the dedicated vault management screens already introduced in Settings.

### Changes
- Added a dedicated Share Personal Vault screen with the updated Figma-based UI.
- Implemented the access code section with QR generation and vault link copy action.
- Added the access permissions section with Viewer, Editor, and Admin options.
- Added the access controls section with Access Once toggle and Access Expires selector.
- Added invite expiry countdown/progress UI and refresh behavior tied to invite regeneration.
- Registered the new share screen in the settings navigator.
- Updated the share navigation flow so Share Personal Vault opens the dedicated screen instead of the previous bottom-sheet flow.
- Added Jest coverage for the new share screen.

### Testing Notes
- Ran:
  - `npm test -- --runInBand src/screens/Settings/VaultShareScreen/index.test.jsx`
  - `npm run lint`
  - `npm run typecheck`
- Verified:
  - invite creation on screen mount
  - QR render from invite public key
  - vault link copy action
  - expiry option cycling
  - back navigation
  - invite cleanup on unmount

- The navigation entry point should match the intended user flow from vault actions.
- The current SDK only exposes invite creation/deletion and `publicKey`, so permissions/access-control selections are currently UI-only and not persisted server-side / SDK-side.
- The share flow now uses a dedicated screen instead of the prior bottom-sheet implementation.

### Screenshots/Recordings
simulator_screenshot_2D10CCC1-BBCE-4DC7-8E69-9805364C2E99
Simulator Screenshot - iPhone 17 - 2026-04-02 at 12 09 19
Simulator Screenshot - iPhone 17 - 2026-04-02 at 12 09 15
Simulator Screenshot - iPhone 17 - 2026-04-01 at 18 17 15

### dependencies
- No blocking code dependencies identified.
- If Home vault actions still need final routing alignment outside the current settings path, that follow-up should be linked here.

---
  - https://app.asana.com/0/0/1213851744313437


boris91 (dismissed):

src/containers/Auth/VaultWizard/StepReview.jsx
```diff
@@ -0,0 +1,251 @@
+import { useLingui } from '@lingui/react/macro'
+import { PROTECTED_VAULT_ENABLED } from '@tetherto/pearpass-lib-constants'
+import { Button } from '@tetherto/pearpass-lib-ui-kit'
+import { colors } from '@tetherto/pearpass-lib-ui-theme-provider/native'
+import { View, Text, StyleSheet, ScrollView } from 'react-native'
+
+export const StepReview = ({ vaultData, onCreate, onBack }) => {
```
ndrkltsk: seems to be dead code. I can't see this component being imported anywhere

src/containers/Auth/VaultWizard/StepSecurity.jsx
```diff
@@ -0,0 +1,488 @@
...
+  ButtonPrimary,
+  ButtonSecondary,
+  InputPasswordPearPass
+} from '../../../libComponents'
+import {
+  getVaultPasswordStrengthMeta,
+  getVaultScreenState
+} from '../../../utils/vaultFlow'
+
+export const StepSecurity = ({ initialData, onNext, onBack }) => {
```
ndrkltsk: seems to be dead code. I can't see this component being imported anywhere

src/screens/Settings/VaultSettingsScreen/index.jsx
```diff
@@ -0,0 +1,607 @@
...
+    navigation.navigate('VaultPasswordScreen', {
+      vaultId,
+      vaultName
+    })
+  }
+
+  const handleShare = () => {
+    collapse?.()
+    navigation.navigate('VaultShareScreen', {
+      vaultId,
```
ndrkltsk: you pass navigation parameters here, but VaultShareScreen doesn't seem to consume those parameters

src/screens/Settings/VaultSettingsScreen/index.jsx
```diff
@@ -0,0 +1,607 @@
...
+      vaultName
+    })
+  }
+
+  const openActionMenu = () => {
+    expand({
+      children: (
+        <BottomSheetVaultAction
+          vaultId={vaultId}
```
ndrkltsk: you removed vaultId from the BottomSheetVaultAction accepted parameters, signature was changed to ({ vaultName, onRename, onPassword, onMembers, onShare, onDelete })

src/utils/passwordPolicy.js
```diff
@@ -0,0 +1,80 @@
...
+  minLength: 'Password must be at least 8 characters long',
+  hasLowerCase: 'Password must contain at least one lowercase letter',
+  hasUpperCase: 'Password must contain at least one uppercase letter',
+  hasNumbers: 'Password must contain at least one number',
+  hasSymbols: 'Password must contain at least one special character'
+}
+
+export const getPasswordValidationMessages = (translate) => ({
+  minLength:
+    translate?.(PASSWORD_VALIDATION_MESSAGES.minLength) ||
```
ndrkltsk: lingui's t is a tagged template literal, so calling it as t('string') works at runtime, but the strings inside PASSWORD_VALIDATION_MESSAGES won't be statically extracted by lingui's compiler.
the previous pattern of inline t`Password must be...` in each component was extraction-safe, we should verify the i18n catalog still picks these up

src/utils/vaultFlow.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  return 'idle'
+}
+
+export const buildVaultAccessEntries = ({ devices = [] } = {}) => [
+  {
+    id: 'owner',
+    kind: 'owner',
+    name: 'You',
```
ndrkltsk: we should wrap this using lingui

src/main.jsx
```diff
@@ -40,23 +42,83 @@ i18n.activate('en')
...
+            backgroundColor: colors.grey500.mode1,
+            paddingHorizontal: 24
+          }}
+        >
+          {initError ? (
+            <>
+              <Text
+                style={{
+                  color: colors.white.mode1,
+                  fontFamily: 'Inter',
```
ndrkltsk: we should avoid using inline styles and stay consistent to the rest of the codebase using `StyleSheet.create`

src/containers/Auth/VaultWizard/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+import { StepIdentity } from './StepIdentity'
+import { NAVIGATION_ROUTES } from '../../../constants/navigation'
+import { TOAST_CONFIG } from '../../../constants/toast'
+import { logger } from '../../../utils/logger'
+
+export const VaultWizard = () => {
+  const { t } = useLingui()
+  const navigation = useNavigation()
+  const [formData, setFormData] = useState({
```
ndrkltsk, line 24: this seems to be a write-only state variable

src/utils/vaultFlow.js
```diff
@@ -0,0 +1,58 @@
+import { getPasswordStrengthMeta } from './passwordPolicy'
+
+export const getVaultPasswordStrengthMeta = (password, errors = {}) => {
```
ndrkltsk: this seems to be just a passthrough, why don't we just use getPasswordStrengthMeta?

src/utils/passwordPolicy.js
alvaropaco: You were right that the previous translate?.('raw string') pattern in passwordPolicy.js was not extraction-safe for Lingui.
I kept the existing runtime fallback strings, but made the utility extraction-safe by adding Lingui message descriptors with msg in passwordPolicy.js

src/main.jsx
alvaropaco: Agreed! I removed the inline styles from main.jsx and moved them into a local StyleSheet.create(...) .

src/containers/Auth/VaultWizard/index.jsx
alvaropaco: formData is intentionally kept in VaultWizard because StepIdentity is unmounted while isLoading is true . We call setFormData(data) right before the async create request, and if creation fails, isLoading goes back to false , StepIdentity remounts, and its local state is re-seeded from initialData={formData} . So this is not write-only state; it preserves the user’s entered values across the temporary loading unmount/remount cycle. If we later refactor the screen to keep StepIdentity mounted and show a loading overlay instead, then this state could be removed.`

src/utils/vaultFlow.js
alvaropaco: I agreed with that review point and removed the passthrough.

src/containers/Auth/VaultWizard/index.jsx
alvaropaco: > this seems to be a write-only state variable

@ndrkltsk let me know your thoughts

src/containers/Modal/ModifyVaultModalContent/index.jsx
```diff
@@ -1,16 +1,20 @@
...
-import { ModifyVaultsModaContentWrapper } from '../ModifyVaultsModaContentWrapper'
-import { InputLabel, InputWrapper } from './styles'
+import { TOAST_CONFIG } from '../../../constants/toast'
 import { VAULT_ACTION } from '../../../constants/vaultActions'
 import { useModal } from '../../../context/ModalContext'
 import { logger } from '../../../utils/logger'
+import { getPasswordStrengthMeta } from '../../../utils/passwordPolicy'
+import { ModifyVaultsModaContentWrapper } from '../ModifyVaultsModaContentWrapper'
 
 export const ModifyVaultModalContent = ({
```
giorgikh93, line 25: @alvaropaco I thought we were creating V2 components for re-design e.g. CreatePasswordScreenV2 and rendering them behind the v2 flag, but in this PR I see that you are modifying existing components, I am wondering is it OK to modify existing components or we need to create the copy with V2 suffix?
@denislupookov what do you think?
I am asking cause I am not quite sure about our strategy around such changes.

src/screens/Settings/VaultRenameScreen/index.jsx
```diff
@@ -0,0 +1,362 @@
...
+      })
+    } finally {
+      setIsLoading(false)
+    }
+  }
+
+  return (
+    <KeyboardAvoidingView
+      style={styles.container}
+      behavior={Platform.OS === 'ios' ? 'padding' : undefined}
```
giorgikh93: can we create a helper for this check ?  I think it would reduce complexity of our codebase

const isIOS = () => Platform.OS === "ios"
const isAndroid=() => Platform.OS === "android"

src/worklet/index.js
```diff
@@ -36,24 +53,43 @@ export const createPearpassVaultClient = async ({ debugMode } = {}) => {
 
   worklet.start('/worklet.bundle', bundle)
 
-  const sharedDirectory = await getSharedDirectoryPath()
+  logger.log('[pearpass] init: worklet started')
+
+  const sharedDirectory = await withTimeout(
```
giorgikh93: Why we need additional timeout for getting shared directory path? Did you encounter any issues?

package.json
```diff
@@ -8,13 +8,15 @@
     "start": "expo start --dev-client",
     "build": "npm run bundle-bare && npm run lingui:extract && npm run lingui:compile",
     "android": "expo run:android",
-    "ios": "expo run:ios",
+    "ios": "npm run barekit:link-ios-addons && npm run pods:install && expo run:ios",
```
giorgikh93: I am just wondering the purpose of the script changes? Is it because our startup approach changed or ?
@ndrkltsk please review, I personally didn't encounter any issues while starting applications, I'd like to make sure changes are necessary for design updates.

src/containers/Modal/ModifyVaultModalContent/index.jsx
alvaropaco: That's fair @giorgikh93 ... what do you think @denislupookov ?

src/worklet/index.js
alvaropaco: Good question. I did not hit a specific failure on getSharedDirectoryPath() itself in this work. The timeout was added as a defensive guard because this is part of the app’s startup-critical path: if the native App Group call ever hangs instead of resolving/rejecting, createPearpassVaultClient() would stall before the vault client is created. Rejections are already handled inside getSharedDirectoryPath() by returning null , so the timeout is only protecting against a never-resolving native call, not normal errors. That said, since I haven’t reproduced an actual hang on this call, I’m happy to remove it if we want to keep the code path simpler and only retain the timeout around setJobStoragePath() , where I was explicitly guarding a native/worklet boundary.`

package.json
alvaropaco: Not because our startup approach changed. These script changes were added to make the iOS native build deterministic after I reproduced the npm run ios failure in the BareKit [CP] Copy XCFrameworks phase. The issue was that react-native-bare-kit expects generated addon XCFrameworks under node_modules/react-native-bare-kit/ios/addons , but those were not always present/up to date before Xcode ran, and the Pods project could also keep stale addon version references until pod install was rerun. So the intent here is build hygiene, not runtime behavior: generate the BareKit iOS addons before launching, then refresh Pods so Xcode references the current addon versions. If we have a better centralized place for that setup, I’m happy to move it there, but the purpose is specifically to prevent the native iOS build from failing with missing/stale BareKit frameworks.
If we guys want, I can revert this change.

src/containers/Modal/ModifyVaultModalContent/index.jsx
denislupookov: we should not modify any existing components. 
@alvaropaco create a new ModifyVaultModalContentV2 and hide under v2 feature flag please.

src/containers/Auth/shared/AuthFlowFormLayout.jsx
```diff
@@ -0,0 +1,150 @@
...
+      Platform.OS === 'android' ? 'humble-nostalgia' : 'Humble Nostalgia',
+    color: colors.white.mode1,
+    textAlign: 'center'
+  },
+  subtitle: {
+    color: colors.white.mode1,
+    textAlign: 'center'
+  },
+  subtitleWithBack: {
+    color: '#BDC3AC',
```
denislupookov: do not use static colors please, import from the lib

src/containers/Auth/shared/AuthFlowFormLayout.jsx
```diff
@@ -0,0 +1,150 @@
...
+    gap: 20
+  },
+  title: {
+    fontFamily:
+      Platform.OS === 'android' ? 'humble-nostalgia' : 'Humble Nostalgia',
+    color: colors.white.mode1,
+    textAlign: 'center'
+  },
+  subtitle: {
+    color: colors.white.mode1,
```
denislupookov: we should not use old colors

src/containers/BottomSheetAddDeviceContent/index.jsx
```diff
@@ -44,62 +60,66 @@ const BottomSheetAddDeviceContentBase = () => {
...
-              </TouchableOpacity>
-              <TouchableOpacity
-                style={styles.tab}
-                activeOpacity={0.7}
-                onPress={() => setActiveTab(TAB.PAIR_ANOTHER)}
-              >
-                <Text style={styles.tabText}>{t`Import vault`}</Text>
-              </TouchableOpacity>
-            </View>
+            showImportTab ? (
```
denislupookov: is this old or new UI ?

src/containers/BottomSheetVaultAction/index.jsx
```diff
@@ -18,55 +29,124 @@ import { useHapticFeedback } from '../../hooks/useHapticFeedback'
...
-    )
+    handleAction()
+    onPassword?.()
   }
 
   return (
     <>
       <View style={styles.header}>
-        <Text style={styles.title}>{t`What would you like to modify?`}</Text>
+        <Text style={styles.eyebrow}>{t`Vault actions`}</Text>
```
denislupookov: remove style

src/containers/BottomSheetVaultAction/index.jsx
```diff
@@ -18,55 +29,124 @@ import { useHapticFeedback } from '../../hooks/useHapticFeedback'
...
+        <Text style={styles.title}>{vaultName}</Text>
```
denislupookov: do not use style prop for lib ui kit components please.

src/screens/Settings/VaultDeleteScreen/index.jsx
```diff
@@ -0,0 +1,243 @@
...
+      bottomOffset: TOAST_CONFIG.BOTTOM_OFFSET
+    })
+  }
+
+  return (
+    <KeyboardAvoidingView
+      style={styles.container}
+      behavior={Platform.OS === 'ios' ? 'padding' : undefined}
+    >
+      <Svg width="100%" height="100%" style={StyleSheet.absoluteFill}>
```
denislupookov: is this an icon ? Could you move it to the ui kit lib

src/screens/Settings/VaultDeleteScreen/index.jsx
```diff
@@ -0,0 +1,243 @@
...
+      <SafeAreaView style={styles.safeArea} edges={['top', 'left', 'right']}>
+        <View style={styles.contentWindow}>
+          <View style={styles.header}>
+            <TouchableOpacity
+              activeOpacity={0.7}
+              onPress={() => navigation.goBack()}
+              style={styles.backButton}
+              testID="vault-delete-back-button"
+            >
+              <BackIcon size={20} color={colors.white?.mode1 || '#FFFFFF'} />
```
denislupookov: pass useTheme color from ui lib kit

src/screens/Settings/VaultDeleteScreen/index.jsx
```diff
@@ -0,0 +1,243 @@
...
+import {
+  Keyboard,
+  KeyboardAvoidingView,
+  Platform,
+  StyleSheet,
+  Text,
+  TextInput,
+  TouchableOpacity,
+  View
+} from 'react-native'
```
denislupookov: use lib ui kit components: Text, Input, Button ...

denislupookov (changes requested):
Could you change everywhere to use `@tetherto/pearpass-lib-ui-kit`: colors, tokens, components, etc...

src/worklet/index.js
giorgikh93: @alvaropaco In my opinion, we should separate design features from technical improvements to avoid introducing additional complexity alongside design changes. I suggest removing any technical-related updates from this PR and if really necessary, creating a separate PR where we can properly review and discuss the pros and cons.

package.json
giorgikh93: @alvaropaco Thanks for your explanation. My point is the same as - https://github.com/tetherto/pearpass-app-mobile/pull/284#discussion_r3042871270, I'd like @ndrkltsk to have a look as well, thanks.

src/containers/BottomSheetAddDeviceContent/index.jsx
alvaropaco: Old... reverted

boris91 (approved):

src/containers/Modal/ModifyVaultModalContent/index.jsx
```diff
@@ -136,50 +201,170 @@ export const ModifyVaultModalContent = ({
...
+  },
+  strengthHeader: {
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    alignItems: 'center'
+  },
+  strengthTrack: {
+    height: 8,
+    borderRadius: 999,
+    backgroundColor: '#202614',
```
denislupookov, line 346: can you remove hardcoded colors please everywhere

denislupookov (approved):

PR #286, opened 2026-04-03
v2 components for Records screens (edit and view)

harrisaar-noxtton:
## prerequisites 

Needs updates from https://github.com/tetherto/pearpass-lib-ui-react-native-components/pull/57

## Task

We must use v2 design components for records screens. View and Edit mode.

Following screens in edit and read mode (14 screens total):
1. Login 
2. Credit card
3. Custom
4. Identities
5. Note
6. Passphrase
7. Wifi

## solution

We created 2 folders for both edit/view mode. Folder `v1` and `v2`. Selection is done based on DESIGN_VERSION constant value.

```
import { DESIGN_VERSION } from '@tetherto/pearpass-lib-constants'

export const RecordDetails = (props) => {
  if (DESIGN_VERSION === 1) return <RecordDetailsV1 {...props} />
  return <RecordDetailsV2 {...props} />
}
```

and

```
export const CreateRecord = (props) => {
  if (DESIGN_VERSION === 1) return <CreateRecordV1 {...props} />
  return <CreateRecordV2 {...props} />
}
```

## Todo components

For the following components, we should create a separate task. Since these are components implemented in the mobile package, they have more functionality than just UI appearance.

1. UploadField - it is implemented in lib-ui-kit, BUT in the mobile package we need to implement file selection and this component has to be re-used in all the screens.

2. OtpCodeField - Located in the 'view' mode of login records screen.

3. PassPhrase - Located in the passphrase records screen page. 

4. ImageField - Used in Identities screen for passport, id photos.

5. PressableNote - Used in the notes screen. (Not sure what it is meant for, looks like regulate InputField)

## Todo related tasks

1. According to the Figma spec, all screens should have the v2 design header, and the save button should be located at the bottom. I noticed a related task for this. Once the new header component has been created, it must be applied to all 14 screens.

2. The background color needs updating — currently gray, but should be greenish per Figma. This PR focuses on v2 component integration and conditional screen selection, so style tweaks are out of scope here, but will likely be addressed alongside the new header task.

## Architectural recommendations

1. The previous architecture maintained separate files for each screen’s two views, which seems unnecessary. The difference between views is simply whether inputs are readonly or editable — this could be handled with a disabled prop on each input, changing the style accordingly. This would significantly reduce the number of files, with the readonly/editable logic condensed to a single line. 

I kept it as is atm, since the changes would be too significant from the original approach.

## Product design recommendations

1. The current 7 password entry screen variations are excessive. Limiting to 3 options would cover most use cases, reduce maintenance overhead, and align with UX best practices around minimizing user choices.

2. Many password entry screens have both “note” and “custom fields” inputs, which is redundant and confusing. Replacing them with a single MultipleSlotInput component would simplify the code and clean up the UI.

## Clicking through "edit" screens

https://github.com/user-attachments/assets/35a6ea80-389d-4036-a860-3ba3d9ff79c3

## Clicking few 'view' screens

https://github.com/user-attachments/assets/89b0fe7b-f144-499c-9e15-232fb24905f5


boris91 (approved):

src/screens/CreateRecord/v2/CreateOrEditCustomContent.tsx
```diff
@@ -0,0 +1,223 @@
...
+              onRemove={(index: number) => removeCustomField(index)}
+              errorMessage={(errors as Record<string, {error?: {note?: string}}[]>)?.customFields?.find(Boolean)?.error?.note}
+              testID="custom-fields-multi-slot-input"
+            />
+          </FormWrapper>
+        </ScrollView>
+      </ScrollContainer>
+    </Wrapper>
+  )
+}
```
alvaropaco, line 223:
```suggestion
}

```

src/screens/CreateRecord/v2/CreateOrEditIdentityContent.tsx
```diff
@@ -0,0 +1,621 @@
...
+      drivingLicenseExpiryDate: initialRecord?.data?.drivingLicenseExpiryDate ?? '',
+      drivingLicenseIssuingCountry: initialRecord?.data?.drivingLicenseIssuingCountry ?? '',
+      drivingLicensePicture: initialRecord?.data?.drivingLicensePicture || [],
+      attachments: initialRecord?.attachments ?? []
+    },
+    validate: (values) => schema.validate(values)
+  })
+
+  useGetMultipleFiles({
+    fieldNames: ['attachments', 'passportPicture', 'idCardPicture', 'drivingLicensePicture'],
```
alvaropaco, line 238:
```suggestion
    fieldNames: fileNamesList,
```

src/screens/CreateRecord/v2/CreateOrEditIdentityContent.tsx
```diff
@@ -0,0 +1,621 @@
...
+      drivingLicenseDateOfIssue: initialRecord?.data?.drivingLicenseDateOfIssue ?? '',
+      drivingLicenseExpiryDate: initialRecord?.data?.drivingLicenseExpiryDate ?? '',
+      drivingLicenseIssuingCountry: initialRecord?.data?.drivingLicenseIssuingCountry ?? '',
+      drivingLicensePicture: initialRecord?.data?.drivingLicensePicture || [],
+      attachments: initialRecord?.attachments ?? []
+    },
+    validate: (values) => schema.validate(values)
+  })
+
+  useGetMultipleFiles({
```
alvaropaco, line 237:
```suggestion
const fileNamesList = ['attachments', 'passportPicture', 'idCardPicture', 'drivingLicensePicture'] as const;
  useGetMultipleFiles({
```

src/screens/CreateRecord/v2/CreateOrEditIdentityContent.tsx
```diff
@@ -0,0 +1,621 @@
...
+      type: 'baseToast',
+      text1: error.message,
+      position: 'bottom',
+      bottomOffset: 100
+    })
+  }
+
+  const onSubmit = async (values) => {
+    if (isLoading) return
+
```
alvaropaco, line 261:
```suggestion

if(Object.keys(values).length === 0) return

```

src/screens/CreateRecord/v2/CreateOrEditWifiPasswordContent.tsx
```diff
@@ -0,0 +1,190 @@
...
+              onRemove={(index: number) => removeCustomField(index)}
+              errorMessage={(errors as Record<string, {error?: {note?: string}}[]>)?.customFields?.find(Boolean)?.error?.note}
+              testID="custom-fields-multi-slot-input"
+            />
+          </FormWrapper>
+        </ScrollView>
+      </ScrollContainer>
+    </Wrapper>
+  )
+}
```
alvaropaco, line 190:
```suggestion
}

```

denislupookov:
Need additional changes to apply new header / layout design not just change the components

giorgikh93:
New PR will be provided by @denislupookov

PR #316, opened 2026-04-16
ci:adding workflows authorize gate + PR-head context

kapildev421:
### Requirements

- Run a CI workflow (lint, unit tests) for every PR opened against main such that:
- The workflow validates the PR head code, not the target branch.
- Triggers on opened, reopened, and synchronize.
- Runs only for trusted contributors (org members, collaborators, or external PRs explicitly opted-in by a maintainer via the verify label) — required because this is a public repo.
- using scherermichael-oss/action-has-permission to check the permission

### Changes

Replaces lint.yml + test.yml with a three-file CI pipeline


torkinos:
@nickgr2 please review

nickgr2 (approved):
The `pull_request_target` design is sound — authorize gate runs base-repo code only, PR-head checkout is gated behind the allow output, `persist-credentials: false` on the PR-head checkout, no `secrets: inherit` on the reusable workflow, and synchronize blocks unconditionally so a still-attached label can't carry over to new commits. Pinning third-party actions by SHA is the right call.

A few notes:

**1. Third-party `scherermichael-oss/action-has-permission` — consider inlining**

It's pinned by SHA so rug-pull is mitigated, but the action is a single API call to `GET /repos/{owner}/{repo}/collaborators/{actor}/permission`. The whole "actor has write" branch could be ~6 lines of bash + `gh api` and remove one supply-chain dependency entirely:

```yaml
- name: Check actor permission
  id: perm
  shell: bash
  env:
    GH_TOKEN: ${{ inputs.github-token }}
    ACTOR: ${{ github.actor }}
  run: |
    level=$(gh api "repos/${{ github.repository }}/collaborators/${ACTOR}/permission" --jq .permission 2>/dev/null || echo "none")
    case "$level" in admin|maintain|write) echo "has-permission=1" >> "$GITHUB_OUTPUT" ;; *) echo "has-permission=0" >> "$GITHUB_OUTPUT" ;; esac
```

The upstream `scherermichael-oss/action-has-permission` repo hasn't been updated in years and is a personal account; trimming the dep surface here seems worthwhile. Non-blocking.

**2. `ci-push.yml` branch filter form**

```yaml
branches: main       # ci-push.yml
branches: [main]     # ci-pr.yml
```

Both are valid; aligning them (use the list form in both) avoids the next reader wondering if the difference is meaningful. Minor.

**3. Silent label-delete failure**

```bash
gh api -X DELETE "repos/${{ github.repository }}/issues/${PR_NUMBER}/labels/${LABEL}" 2>/dev/null || true
```

Defense-in-depth is fine here (the synchronize branch in the check step already rejects unconditionally so a failed strip doesn't bypass the gate), but swallowing stderr means a real failure — token scope, API rate limit, etc. — goes unnoticed and the next maintainer can't see why the label kept coming back. Suggest dropping `2>/dev/null` and keeping `|| true`, so the failure surfaces in the log without failing the job. Minor.

**4. Inherent trust model of the `verify` label**

Worth documenting somewhere (PR description, CONTRIBUTING, or a comment in the workflow): when a maintainer applies `verify`, the next CI run executes arbitrary PR code via `npm ci` install scripts + `npm test`. Reviewers should treat the label as a "I've eyeballed the diff including any dep changes" sign-off, not just a "looks legit." Not a code change, just a recommendation.

kapildev421:
@nickgr2  please review

torkinos (approved):

hakaari (approved):

PR #320, opened 2026-04-17
Dev to main

torkinos:


boris91 (dismissed):

app.json
```diff
@@ -3,7 +3,7 @@
     "name": "PearPass",
     "slug": "pearpass",
     "icon": "./assets/images/appIcons/app-icon.png",
-    "version": "1.6.1",
+    "version": "1.6.0",
```
giorgikh93: @torkinos I guess 1.6.1 should be an actual version

torkinos: good catch, thanks!

boris91 (approved):

giorgikh93 (approved):

PR #321, opened 2026-04-17
feat: implementing autofill extension V2 desing

alvaropaco:
### Requirements
- V2 autofill redesign: implemented a native SwiftUI design system inside `PearPassAutofillExtension` that mirrors the `pearpass-lib-ui-react-native-components` kit.
- The kit itself (React-Strict-DOM / StyleX) can't be consumed by the extension — it runs outside the RN runtime — so tokens and components have to be re-implemented natively.
- Ship dark theme only for now. Upstream `light.ts` is intentionally empty; light support can come later.
- No screen-level wiring in this PR. These are primitives, available for the upcoming screen rebuilds to consume.

### Changes
Added a new `DesignSystem/` folder under the iOS extension with the full set of v2 primitives:

- **`Tokens.swift`** — Spacing, Radius, FontFamily, FontSize, FontWeight. One-to-one mirror of `src/theme/tokens.raw.ts`.
- **`Colors.swift`** — 27-property `ThemeColors` struct + `.dark` palette, sourced from `src/theme/themes/dark.ts`.
- **`Theme.swift`** — `@Environment(\.pearpassTheme)` + `.pearpassTheme()` modifier + `Font.pearpass(size:weight:)` / `Font.pearpassDisplay(size:)` helpers.
- **Eight `PP*` components**: `PPText` (5 variants), `PPTitle`, `PPButton` (4 variants × 2 sizes, icon slots, loading/disabled), `PPInputField` (label + error + generic left/right slots), `PPPasswordField`, `PPSearchField`, `PPListItem` (single/split subtitle, destructive, divider), `PPPageHeader`.

Also modified `plugins/expo-autofill-plugin/src/ios/withIosAutofillExtension.ts` to inject `SWIFT_OPTIMIZATION_LEVEL = "-Onone"` and `SWIFT_COMPILATION_MODE = singlefile` into the extension target's **Debug** config only. Without this, SwiftUI Previews fail with "Cannot preview in this file — Not built with -Onone". Release config is untouched.

### Testing Notes
- Each component ships a SwiftUI `#Preview` covering its relevant states (variants, sizes, error, disabled, loading, etc.) — open any `PP*.swift` in Xcode and Resume the Canvas.
- Verified `xcodebuild -scheme PearPass -configuration Debug` and `-scheme PearPassAutoFillExtension` both build clean against the iOS 15.1 SDK.
- No runtime behavior touched — V1 XML/Fragments and the current iOS extension flow are unaffected.

- **Dark-theme-only** is intentional, not an oversight. `ThemeColors.dark` is the single palette; a light palette lands when upstream fills one in.
- **`PPTitle` falls back to `FontFamily.Default`** — Humble Nostalgia is not bundled in the extension yet. Flagged as a TODO in the source. Renders as system font for now.
- **Debug-only build settings**: new `-Onone` / `singlefile` injection is gated on the Debug config by name. The loop in `withIosAutofillExtension.ts` now discriminates configs by name — worth a quick sanity check that Release is untouched.
- **Generic slot pattern** in `PPInputField<LeftSlot: View, RightSlot: View>` / `PPListItem<Leading: View, Trailing: View>` follows SwiftUI idioms rather than type-erased `AnyView` — slightly more verbose call sites but zero allocation overhead.
- **Naming**: all components are `PP`-prefixed to avoid collisions with SwiftUI built-ins (`Text`, `Button`, `TextField`) and to signal "PearPass design system" at call sites.

### Screenshots/Recordings

https://github.com/user-attachments/assets/76cede45-cab2-4e1f-94d5-5a7622f6d210

### dependencies
- **Blocks on**: V2 feature-flag native bridge (main-app native module writes the flag to App Group / SharedPrefs so the extension can select V1 vs V2 roots at runtime). Follow-up PR.
- **Upstream UI kit**: `pearpass-lib-ui-react-native-components` — source of truth for tokens and color values.


boris91 (dismissed):

torkinos (changes requested):
Thanks @alvaropaco, could you please also add recording of updated UI

boris91 (approved):

ndrkltsk (changes requested):
the attached video shows quite a different UI compared of what we have on figma

PR #331, opened 2026-04-27
Feat/ ios autofill v2

nikushaomanidze:
### Changes

 - Inter and Humble Nostalgia fonts (registered via UIAppFonts in extension Info.plist)               
  - design tokens: colors, spacing, radii, font sizes, typography                                      
  - 10 SVG vector icons in PPV2Assets.xcassets                                                         
  - PPButton with primary / secondary / tertiary / destructive variants                                
  - input / password / search / list-item / content-card / vault-picker primitives                     
  - step number + credential icon backgrounds                                                          
  - credential row + vault inline item row                                                             
  - translucent V2 sheet container (UISheetPresentationController, 0.85 fraction detent, flat corners) 
  - sheet header includes (logo + close, back + close)                                                 
  - V2 screen layouts: loading, master password, combined items, passkey form, error boundary, missing 
  configuration                                                                                        
  - design version feature flag — withIosDesignVersion plugin reads MOBILE_DESIGN_VERSION and writes   
  PEARPASS_DESIGN_VERSION into the extension's Info.plist at prebuild                                  
  - pbxproj patch plugin (withIosExtensionExceptions) — adds membershipExceptions for Info.plist +   
  extension.bundle so the synchronized root group doesn't double-produce them                          
  - V2HostView + V2SheetPresenter + CredentialProviderViewController V1/V2 routing (assertion +      
  registration paths)                                                                                  
  - full passkey save flow ported from V1 PasskeyRegistrationView (PasskeyJobCreator +               
  AuthenticatorDataBuilder + PasskeyCrypto integration)         
  
### Testing Notes

IOS simulator - 26

### Screenshots/Recordings

https://github.com/user-attachments/assets/685eef83-6339-4c85-a239-45954b456399


hakaari:
@nickgr2 can you please review?

nickgr2 (changes requested):
Native iOS V2 design layer for the autofill extension — Swift/SwiftUI port of the existing Android V2. UI-kit rules (`@tetherto/pearpass-lib-ui-kit`, Lingui, `testID`) are RN-specific so they don't apply to extension code; review focused on correctness, localization parity with V1, dead code, and the prebuild plugins. A handful of issues to address before merge.

### Required changes

**`plugins/expo-autofill-plugin/src/ios/withIosDesignVersion.ts:54-58`** — the plugin writes `PEARPASS_DESIGN_VERSION` into the checked-in template at `ios-template/PearPassAutofillExtension/Info.plist`, not just into the prebuild output. Every `expo prebuild` run mutates a tracked file in the repo, which produces uncommitted-changes drift in CI and fresh checkouts and means whoever last ran prebuild has flipped the committed default. The plugin should only write to the generated plist under `cfg.modRequest.platformProjectRoot/PearPassAutofillExtension/Info.plist` (the second branch already does this). If the goal is "write before `withIosAutofillExtension` copies the template," reorder the plugins in `src/index.ts` instead of mutating sources. Drop the template-write path.

**Hardcoded English strings throughout V2 views** — V1 has a populated `Base.lproj/Localizable.strings` and uses it for "Continue", "Go Back", "Open the PearPass app", etc. V2 reintroduces literals: `"Sign in"`, `"Create Passkey"`, `"Enter Your Master Password"`, `"Please enter your master password to continue"`, `"Continue"`, `"Save & Add Login"`, `"Discard"`, `"Go Back"`, `"Add New Login"`, `"Search in All Items"`, `"Select Vault"`, `"No matching items in this vault"`, `"We couldn't find a matching item in this vault…"`, `"save it as a new item"`, the three setup steps, `"Toggle password visibility"`, `"Close"`, `"Back"`, accessibility labels, etc. Wrap them in `NSLocalizedString` (or `String(localized:)`) and add the new keys to `Base.lproj/Localizable.strings` so the V2 path stays translatable like V1.

**`PasskeyFormV2View` doesn't reflect save-in-flight state** — `V2HostView.handleFormSave` sets `isSavingPasskey = true`, but `PasskeyFormV2View` has no `isSaving` prop and the Save button never disables. With async save (passkey crypto + job creation), a user can tap "Save & Add Login" twice and queue two registrations. Add `var isSaving: Bool = false` to `PasskeyFormV2View`, pass `isSavingPasskey`, and either `.disabled(isSaving)` the Save button or swap label/spinner.

### Dead code to remove (or wire up)

- **`V2/V2SheetPresenter.swift`** — defines a `configure(_:)` helper for the 0.85 detent + flat corners but is never called. The PR description says V2 uses `UISheetPresentationController` with a 0.85 fraction detent; the actual code in `attachV2HostView` only sets `preferredCornerRadius = 0` and the 85% height is enforced via SwiftUI in `PPSheetContainer.frame(height: proxy.size.height * 0.85)`. Either invoke `V2SheetPresenter.configure(self)` from `attachV2HostView` (and update `PPSheetContainer` to stop double-fixing the height) or delete the file.

- **`V2/Components/PPSheetContainer.swift`** — defined but no view uses it. `LoadingV2View`, `MasterPasswordV2View`, `CombinedItemsV2View`, `PasskeyFormV2View`, `ErrorBoundaryV2View`, `MissingConfigurationV2View` all wrap content in `VStack { PPSheetHeader… ; PPContentCard … }` directly. Either route the V2 screens through `PPSheetContainer` or drop the file.

- **`V2HostView.loadedFolders`** (`V2HostView.swift:1647, 1822`) — populated from `client.listFolders()` and never read. `PasskeyFormV2View.onSelectFolder: {}` is also a no-op and `folderName: nil` is hardcoded, so the "Select folder" row in the form is inert. If folder picking is out of scope for this PR, remove `loadedFolders`, the no-op closure, and the `PPListItem(title: folderName ?? "Select folder", …)` row so the form doesn't ship UI affordances that don't do anything.

- **`CombinedItemsV2View` vault-password state** — `requiresVaultPassword: Bool`, `@Binding var vaultPassword`, `onUnlockVault: () -> Void` and the commented-out 30-line `vaultPasswordPrompt` (`CombinedItemsV2View.swift` lines 2347–2378 in the diff) are unreachable. The comment says "business logic does not support per-vault passwords yet" — drop the props, the call-site bindings (`requiresVaultPassword: false`, `vaultPassword: bindingFor(\.vaultPassword)`, `onUnlockVault: {}` in `V2HostView`), and the commented block. Re-add when the feature lands.

- **`PasskeyFormV2View.titleError` / `websiteError`** props — declared but never set by the caller (only `formSaveError` is wired). `handleFormSave` writes "Title is required" into `formSaveError` instead. Either route inline title/url validation through these props or drop them.

### Minor

- `MasterPasswordV2View.swift` `Spacer().frame(height: 46)`, `MissingConfigurationV2View.swift` `HStack(spacing: 10)`, `PPCredentialItemView.swift` `VStack(spacing: 2)` — magic numbers in new V2 code. `PPSpacing.s10` already exists; pull the rest into the spacing scale (or add `s46` if the figma calls for it) so the tokens stay the source of truth, since that's the whole point of `PPSpacing`.
- `V2HostView.extractDomain` doesn't strip a leading `www.`, so a credential saved at `www.google.com` and an assertion targeting `google.com` won't match each other (only the suffix-match in one direction holds). Worth normalizing both sides.

nikushaomanidze:
@nickgr2 added new commit. can you please review of this pr again?

nickgr2 (changes requested):
Re-review against `e8e6d4b`. Three new commits since `a3dabc5`: V1-parity work in `V2HostView` (registration UPDATE_PASSKEY flow, pending-jobs merge into the loaded list, vault-unlock fallback ladder, search filter, sticky `hasUserSearched`, www-normalization), passing `passkeyRpId` into the host view, and flipping `PEARPASS_DESIGN_VERSION` to 2.

The new V1-parity logic looks reasonable (UPDATE vs ADD branch, pending-job merge, www-strip). However, most of the **required** changes from the prior review are still not addressed.

### Still unresolved from prior review

- **`plugins/expo-autofill-plugin/src/ios/withIosDesignVersion.ts:63-75`** — the plugin still writes into the checked-in template at `ios-template/PearPassAutofillExtension/Info.plist`. This commit flipping the tracked plist to `<integer>2</integer>` is itself a symptom of the issue: any contributor who runs `expo prebuild` mutates a tracked file. Drop the template-write branch and rely on the second branch that writes to the generated plist (or reorder plugins in `src/index.ts` if the goal is to write before the template copy).
- **Hardcoded English strings throughout V2 views** — still no `NSLocalizedString` / `String(localized:)` anywhere in `V2/Views/*.swift`. "Sign in", "Create Passkey", "Enter Your Master Password", "Continue", "Save & Add Login", "Discard", "Go Back", "Add New Login", "Search in All Items", "Select Vault", "No matching items in this vault", "save it as a new item", the three setup steps, "Toggle password visibility", "Close", "Back", and accessibility labels still ship as bare literals. V1 routes these through `Base.lproj/Localizable.strings`; V2 needs the same treatment to keep the `PEARPASS_DESIGN_VERSION=2` path translatable.
- **`V2/Views/PasskeyFormV2View.swift:11-30`** — still no `isSaving` prop. `V2HostView.handleFormSave` sets `isSavingPasskey = true` but the form's Save button never disables. With the new UPDATE/ADD branching the failure mode is worse — a double-tap can fan out across two job-creation paths. Add `var isSaving: Bool = false`, pass `isSavingPasskey`, and `.disabled(isSaving)` the Save button (or swap label/spinner).
- **`V2/V2SheetPresenter.swift`** — still defined, still never called. The actual code in `attachV2HostView` only sets `preferredCornerRadius = 0`; the 0.85 detent is still enforced via SwiftUI in `PPSheetContainer.frame(...)` (which is also dead — see next). Either invoke `V2SheetPresenter.configure(self)` or delete the file.
- **`V2/Components/PPSheetContainer.swift`** — still defined, still not imported by any view. Each V2 screen wraps content in `VStack { PPSheetHeader… ; PPContentCard … }` directly. Either route screens through it or delete.
- **`V2/Views/CombinedItemsV2View.swift:41-48, 251-277`** — `requiresVaultPassword`, `vaultPassword` binding, `onUnlockVault`, and the `vaultPasswordPrompt` view body are still unreachable dead code. Call site still passes `requiresVaultPassword: false`. Drop them; re-add when per-vault password lands.
- **`V2/Views/PasskeyFormV2View.swift:22-23`** — `titleError` and `websiteError` props still declared, still never set by the caller (only `formSaveError`/`saveError` is wired). Title required is still surfaced via `formSaveError = "Title is required"` instead of `titleError`. Either route inline title/url validation through these or drop them.
- **`loadedFolders` / `onSelectFolder` / `folderName: nil`** — the host still loads folders into `V2HostView.loadedFolders` (`V2HostView.swift:50, 244`) but the form is still called with `folderName: nil, onSelectFolder: {}` (lines 170, 174). The "Select folder" `PPListItem` row in `PasskeyFormV2View` is therefore still inert. The new save flow does preserve `selectedExistingRecord?.folder`, which is good, but the no-op picker UI shouldn't ship — drop `loadedFolders`, the closure, and the row.
- **Magic spacing** — `MasterPasswordV2View.swift:46` `Spacer().frame(height: 46)`, `MissingConfigurationV2View.swift:71` `HStack(spacing: 10)`, `PPCredentialItemView.swift:27` `VStack(alignment: .leading, spacing: 2)` are still raw numbers. Pull through `PPSpacing` (`s10` already exists; add `s2`/`s46` if the figma calls for them).

### Resolved

- `V2HostView.extractDomain` now strips `www.` and lowercases — domain comparisons survive www-prefixed websites. ✅

### One concern with the new code

- **`V2HostView.activateVault`** (`V2HostView.swift:498-519`): the `decryptVaultKey` call falls back to `""` for missing `ciphertext`/`nonce`:
  ```swift
  ciphertext: masterEncryption["ciphertext"] as? String ?? "",
  nonce: masterEncryption["nonce"] as? String ?? "",
  ```
  If the master encryption record is malformed, this hands empty strings to crypto and the failure surfaces as "decryptionFailed" rather than the actual missing-field issue. Throw a clear `PearPassVaultError` upfront when either field is missing — easier to diagnose in field reports.

Holding at REQUEST_CHANGES until at least the plugin template-write, hardcoded strings, isSaving prop, and dead-code items land.

nikushaomanidze:
@nickgr2 resolve your comments, please review

nickgr2 (approved):
Re-review against `e1c2470`. Six new commits since `e8e6d4b` — each maps to one of the required-changes items from the prior review. Spot-checked the diffs and current file state; all the requested fixes are in place.

### Resolved

- **Plugin template-write** (`b7afe4b`) — `withIosDesignVersion.ts` now writes only to the generated extension plist under `cfg.modRequest.platformProjectRoot`. The template-write branch is gone and the stale `PEARPASS_DESIGN_VERSION` key is dropped from the tracked template plist. Fresh checkouts and CI will no longer drift on prebuild. ✅
- **Hardcoded V2 strings** (`8df3d7e`) — `NSLocalizedString` (or `String(localized:)`) wraps headers, buttons, placeholders, empty-state copy, save-error strings, and accessibility labels across `V2/Views/*.swift`, `V2/Components/*.swift`, and `V2HostView.swift`. New keys grouped in `Base.lproj/Localizable.strings`; existing V1 keys reused. The "Title is required" inline error in `handleFormSave` is also localized. ✅
- **`isSaving` prop on `PasskeyFormV2View`** (`d843bb4`) — prop added, wired from `V2HostView` (`isSaving: isSavingPasskey`), and both Save and Discard `PPButton`s now `.isEnabled(!isSaving)`; Save label swaps to "Saving…". Double-tap fan-out across the new UPDATE/ADD branches is closed off. ✅
- **`V2SheetPresenter.swift` / `PPSheetContainer.swift`** (`9c15b66`) — both deleted. ✅
- **`CombinedItemsV2View` vault-password dead code** (`9c15b66`) — `requiresVaultPassword`, `vaultPassword`, `onUnlockVault` props and the commented-out `vaultPasswordPrompt` removed; matching call-site bindings dropped from `V2HostView`. ✅
- **`PasskeyFormV2View.titleError` / `websiteError` dead props** (`9c15b66`) — dropped along with their inline error UI, since only `saveError` is wired. ✅
- **`loadedFolders` / `onSelectFolder` / `folderName: nil`** (`9c15b66`) — `loadedFolders` state and the `listFolders()` fetch in `handleMasterPasswordContinue` removed; the `folderName` / `onSelectFolder` props and the inert "Select folder" `PPListItem` row removed from the form and the call site. ✅
- **Magic spacing** (`eb67b0c`) — `PPSpacing` gains `s2` and `s46`; `MasterPasswordV2View`, `MissingConfigurationV2View`, and `PPCredentialItemView` now reference `PPSpacing.*` instead of raw numbers. ✅
- **Symmetric `targetDomain` normalization** (`e1c2470`) — `targetDomain` now routes the registration `rpId`, the assertion `passkeyRpId`, and the password-assertion `serviceIdentifiers` path through `extractDomain`, so all three normalize the same way as record sites (lowercased, scheme/path/port/`www.` stripped). Closes the asymmetry the prior review flagged on the assertion side. ✅

### Minor (non-blocking)

- **`withIosDesignVersion.ts` inline comment** (`b7afe4b`) — the comment in the plugin says "withIosAutofillExtension is registered before this plugin (see src/index.ts)", but in `src/index.ts` it is now registered **last**. The `index.ts` comment explains the intended invariant — Expo dangerous mods compose such that the last-registered plugin runs first, so the template copy lands before this plugin runs — but the wording in `withIosDesignVersion.ts` reads the opposite. Worth either rewording the in-plugin comment to match (registered last → runs first) or, defensively, doing a `expo prebuild --clean` from a fresh checkout and confirming `<key>PEARPASS_DESIGN_VERSION</key><integer>2</integer>` lands in `ios/PearPassAutofillExtension/Info.plist` so we know the chain order is what we expect.
- **`activateVault` empty-string fallback** (`V2HostView.swift:526-529`) — unchanged from the prior review, restating for the record. Missing `ciphertext`/`nonce` still get coerced to `""` and surface as `decryptionFailed` rather than a clear "missing field" error. Throw a `PearPassVaultError` upfront when either field is absent; field reports will be much easier to read. Not blocking.

Approving — required changes are in. Once branch-protection signs off and the draft flag is removed, this is good to land.

nikushaomanidze:
@nickgr2 resolve your Minor comments. please review again and approve pr

nickgr2 (approved):
Both minor notes from the prior review addressed in 07643a6:

- `withIosDesignVersion.ts` comment now correctly explains that Expo's dangerous mods compose in reverse registration order (last registered runs first), and the explanation is consistent with `src/index.ts` where `withIosAutofillExtension` is registered last in `withPlugins` so its template-copy mod runs before this plugin.
- `activateVault` replaces the silent empty-string fallback for `ciphertext` / `nonce` with explicit `PearPassVaultError.unknown` guards, matching the existing `hashedPassword` guard above.

LGTM.

nickgr2:
> please review again and approve pr

Already approved at 07643a6 (current head) on 2026-04-28T09:48Z — that approval covers the minor-note fixes.

boris91 (dismissed):

boris91 (approved):

torkinos (approved):

PR #334, opened 2026-04-28
Fix Onboarding/Creation MP screens v2 design

denislupookov:
### Changes

[v2][Android/iOS][Onboarding screens] Bugs
[v2][Android/iOS]["Create master password" screen] Bugs

### Screenshots/Recordings

IOS:
https://github.com/user-attachments/assets/546f47b2-98f5-435f-9bff-0d3f76ce5883

Android:
https://github.com/user-attachments/assets/ef65b92e-7204-4b9b-a025-683c21c62c7b

---
  - https://app.asana.com/0/0/1214114208642587
  - https://app.asana.com/0/0/1214114242643823


giorgikh93:
@nickgr2 please review this PR

denislupookov:
@nickgr2 please review this PR

torkinos:
@nickgr2 please review

shavtvalishvili:
Looks good, test check is failing though.

nickgr2 (changes requested):
Most of the layout/spacing fixes look good, but a few items need addressing before approval.

### Required

1. **i18n regression — `CreatePasswordScreen.jsx:62`**
   The previous subtitle was `${t`This is the key to access PearPass. Already using PearPass?`}`. The refactor dropped the second sentence out of the `t``` wrapper:
   ```jsx
   {t`This is the key to access PearPass.`}{' '}
   {unsupportedFeaturesEnabled() ? (
     <>
       {'Already using PearPass?'}   // ← hardcoded English
   ```
   Wrap it: `{t`Already using PearPass?`}`.

2. **Hardcoded spacing literals where tokens exist**
   - `AuthFlowFormLayout.jsx` `content.paddingTop/paddingBottom: 60`, `contentKeyboardOpen.paddingTop/paddingBottom: 40`
   - `MasterPasswordScreen.jsx` `topSection.paddingTop` was `rawTokens.spacing40` and is now literal `60`

   `rawTokens` exposes `spacing40` (used elsewhere in the repo) — the `40` values should use it. `60` isn't in the token set; per the styling rule (`AGENTS.md` / review guidelines), if the design needs a value not in the kit, raise it with the team to add a token rather than hardcoding. At minimum, please replace the `40`s with `rawTokens.spacing40` and call out the `60` decision in the PR description so it doesn't drift further.

### Questions

3. **Scope** — `plugins/withAppConfig/src/android/withMainActivityConfigChanges.ts` (Android Activity `configChanges` for fontScale/density/fontWeightAdjustment) doesn't appear to relate to the onboarding/create-MP screen design fixes named in the title. Could you split it into its own PR, or add a line to the description explaining why it ships together (e.g. fixes a regression observed on these screens)?

4. **Terms-of-use link change — `CreatePasswordScreen.jsx:86-88`**
   ```jsx
   <Link href={TERMS_OF_USE} isExternal onClick={() => Keyboard.dismiss()}>
   ```
   was replaced with
   ```jsx
   <Link onClick={() => { Keyboard.dismiss(); Linking.openURL(TERMS_OF_USE) }}>
   ```
   Was the kit's `href + isExternal` path broken for this case? If so, that's worth a kit-side fix rather than per-call-site `Linking.openURL` — otherwise other external links in the app will keep using the kit while this one diverges.

### Minor

5. `ConfirmablePasswordFields.jsx`: `style={{ gap: rawTokens.spacing12 }}` — prefer a `StyleSheet.create` entry for consistency with the rest of the file.
6. `CreatePasswordScreen.test.jsx`: `rawTokens: new Proxy({}, { get: () => 0 })` is a clever shim, but it now silently swallows typos (any `rawTokens.foo` returns 0). Returning a sentinel that surfaces in snapshots, or listing the few keys you use, would catch future regressions.

Once 1 and 2 are addressed I'm happy to re-review.

denislupookov:
@nickgr2 re review please

nickgr2 (changes requested):
Thanks for the round of fixes. The i18n wrap, terms-of-use revert, and `ConfirmablePasswordFields` `StyleSheet` move all look good. One blocker on the spacing token though.

### Required

**`rawTokens.spacing60` is not exported by the kit.** I checked `node_modules/@tetherto/pearpass-lib-ui-kit/src/theme/tokens.raw.ts` at the head SHA — `rawTokens` only exposes `spacing2/4/6/8/10/12/16/20/24/32/40/48`. There is no `spacing60`. So the three references in this PR resolve to `undefined`:

- `src/containers/Auth/shared/AuthFlowFormLayout.jsx:147-148` (`content.paddingTop` / `paddingBottom`)
- `src/screens/AuthV2/MasterPasswordScreen.jsx:256` (`topSection.paddingTop`)

In React Native, `paddingTop: undefined` falls back to `0`, so the screens will render with no top/bottom padding instead of 60 — the exact thing this PR is trying to fix gets silently regressed. To verify locally: log `rawTokens.spacing60` (it'll be `undefined`) or run the app and observe the missing padding.

The right fix is to add `spacing60: 60` (and `spacing70` if your designs need it) to `tokens.raw.ts` (and `tokens.json`) in `pearpass-lib-ui-kit`, ship that as a kit release, then bump it here. If you want to land this PR before the kit change, keep the literal `60` and add a one-line PR description note explaining the token is pending — but please don't merge with `rawTokens.spacing60` references that don't resolve.

### Still open from previous review (non-blocking, your call)

- Q3 — Android `withMainActivityConfigChanges.ts` scope. Up to you whether to split or note the connection in the description.
- Minor #6 — the `Proxy({}, { get: () => 0 })` in `CreatePasswordScreen.test.jsx` still swallows typos.

Once `spacing60` resolves to a real value I'm happy to approve.

nickgr2:
@denislupookov Looking at the PR, the head SHA is still `cacece0e` — the same one I reviewed at 11:11 UTC. No new commits have been pushed since my `REQUEST_CHANGES` review.

The blocker hasn't moved: `rawTokens.spacing60` is still not exported by `@tetherto/pearpass-lib-ui-kit` (re-checked `node_modules/.../tokens.raw.ts` at this SHA — only `spacing2/4/6/8/10/12/16/20/24/32/40/48` exist), so `AuthFlowFormLayout.jsx:147-148` and `MasterPasswordScreen.jsx:256` still resolve to `undefined` → `paddingTop/Bottom: 0`.

Could you push the fix (either add `spacing60` to the kit and bump the dep, or revert to literal `60` with a PR-description note) and I'll re-review on the new SHA?

denislupookov:
@nickgr2 double check once again node_modules/@tetherto/pearpass-lib-ui-kit/src/theme/tokens.raw.ts. It has spacing: 60.

> Still open from previous review (non-blocking, your call)

Let's skip it

denislupookov:
@nickgr2 you are not correct. The head SHA was changed to 5f30c6ae68434d8a7a14bc87985d74dd44981c2b in this commit - https://github.com/tetherto/pearpass-app-mobile/pull/334/commits/cacece0efb517b2b826f190a46c6e56f5c1ed73f

nickgr2:
> double check once again node_modules/@tetherto/pearpass-lib-ui-kit/src/theme/tokens.raw.ts. It has spacing: 60.

You're right — verified at the pinned SHA `5f30c6ae` (resolved in `package-lock.json`); `spacing60: 60` is present. My local `node_modules` was stale from before that kit commit, so my re-review was wrong. Sorry for the noise.

> Let's skip it

Agreed, dropping it. Will re-review.

nickgr2 (approved):
Re-reviewed at `cacece0e`. Verified `rawTokens.spacing60` is exported from the kit at the pinned SHA `5f30c6ae` (`src/theme/tokens.raw.ts` lines up to `spacing60: 60`) — my prior REQUEST_CHANGES at this SHA was based on a stale local `node_modules`. Apologies for the noise.

Skipping the non-blocking item per your call. Also confirmed `AlertMessage` is exported from the kit at `5f30c6ae` and its `size='small'` variant intentionally omits the title at render time, so the call site without `title` is safe.

Approving.

giorgikh93 (approved):

torkinos (approved):

PR #337, opened 2026-04-29
[v2] the "current vault" experience

giorgikh93:
### Requirements

Header copy update: Change "Your Vaults" section text to: "Manage your vaults. Select the vault you want to apply changes to."
UI Constraint
Display action icons (Sharing and 3-dots menu) only for the currently active vault in both the navbar dropdown (left side) and in the settings → "your vaults" list
Hide all action icons for inactive vaults
Interaction: Ensure all inactive vault rows are fully clickable.
Navigation: Clicking an inactive vault must switch the active context to that vault
Experience: The vault switch must be seamless,  the user must remain on the current page/view without a reload or redirect, but the new "current vault" should be reflected to the selected one
Validation: Confirm the new active vault now displays the Sharing and 3-dots menu, while the previous vault hides them.
### Changes

- Removed three dots menu from inactive vaults
- Made inactive vaults selectable
- Added unit tests
- Updated "your vaults" description
### Screenshots/Recordings

https://github.com/user-attachments/assets/4a70bc64-fcfa-4f8e-a580-b5c656458141

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,6 +2,7 @@ export default {
   preset: 'react-native',
   transform: { '^.+\\.[jt]sx?$': 'babel-jest' },
   moduleNameMapper: {
+    '^src/(.*)$': '<rootDir>/src/$1',
     '^@tetherto/pearpass-lib-ui-theme-provider/native$':
       '<rootDir>/node_modules/@tetherto/pearpass-lib-ui-theme-provider/native/index.js',
     '^@tetherto/pearpass-lib-ui-theme-provider$':
--- a/src/containers/BottomSheetVaultSelectorContent/index.jsx
+++ b/src/containers/BottomSheetVaultSelectorContent/index.jsx
@@ -1,5 +1,6 @@
 import { useState } from 'react'
 
+import { useBottomSheetModal } from '@gorhom/bottom-sheet'
 import { useLingui } from '@lingui/react/macro'
 import {
   Button,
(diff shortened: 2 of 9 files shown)


giorgikh93:
@nickgr2 please review!

boris91 (dismissed):

nickgr2 (commented):
Refactor and tests look clean. One behavioral question and a couple of minor notes.

### Possible regression: tapping the active vault row in the bottom sheet selector

In `src/containers/BottomSheetVaultSelectorContent/index.jsx`, the previous `handleVaultPress` had a same-id branch that called `closeSelector()`:

```js
if (vault.id === activeVault?.id) {
  closeSelector()
  return
}
```

After the refactor, the active row still wires `onClick={() => switchVault(vault)}` (line 148), but `useVaultSwitch` is invoked without an `onSameVault` callback (line 46-49). So `switchVault` for the active vault now hits the early `onSameVault?.()` no-op and the sheet stays open. Previously a tap on the active row dismissed the sheet.

Is the no-op intentional (i.e. the only meaningful action on the active row is the overflow menu, and dismiss is via the header X)? If so, consider dropping `onClick` for the active row entirely so the intent is explicit. If not, pass `onSameVault: closeSelector` to `useVaultSwitch`.

This is also not covered by the new tests — `index.test.jsx` only asserts `switchVault` is called for the *other* vault row.

### Minor

- `src/hooks/useVaultSwitch.test.js:708-714` — the `useGlobalLoading` assertion only checks the initial call (`isLoading: false`). The protected-vault path toggles `isLoading` true → false → true → false across the modal submit; a follow-up assertion on those transitions would lock in the loading-state behavior the hook is supposed to guarantee.
- `src/screens/Settings/Vaults2/VaultRow.jsx:28` — `onClick={() => onClick?.()}` can be `onClick={onClick}` since `ListItem`'s `onClick` is optional already, but this is style only.

denislupookov (approved):

giorgikh93:
@nickgr2 please review I addressed your points. 

> meaningful action on the active row is the overflow menu, and dismiss is via the header X

 is intentional

nickgr2 (approved):
All three prior comments addressed:

- Active row regression: `onClick={isSelected ? undefined : () => switchVault(vault)}` makes the no-op explicit, with a test asserting the active row's onClick is omitted and `switchVault` is not called. Adding the PersonAdd action on the active row also matches the PR's UI requirement.
- `useVaultSwitch.test.js`: the new test threads call-count tracking through switch + modal submit and asserts terminal idle state, with a comment explaining why React 18 batching can hide intermediate `true` renders. Reasonable.
- `VaultRow.jsx`: `onClick={onClick}` simplification applied.

No new concerns.

boris91 (approved):

torkinos (approved):

PR #339, opened 2026-04-29
Import Vault bug fixes

nikushaomanidze:
### Changes

fixed :
[v2][Android/iOS]["Import Vault" screen] There is a "Copy" icon in the "Vault Link" field
[v2][Android/iOS]["Import Vault" screen] There is redundant "Item" word in the "Item/Vault Link" hint over the "Link" field
[v2][Android/iOS]["Import Vault" screen] The "Vault Link" field is not autofilled out when scanning scanning the QR code
[v2][Android/iOS]["Import Vault" screen] The number of imported items within a Vault is not displayed on the "Import Vault" screen after successful Vault import

### Testing Notes

Android emulator & IOS simulato
### Screenshots/Recordings
| iPhone 17 Pro | Android |
|---|---|
| iPhone Screenshot 1 | Android Screenshot 1 |
| iPhone Screenshot 2 | Android Screenshot 2 |

### Recording

https://github.com/user-attachments/assets/c8d65d70-7cca-449f-94a5-b78dbe0fca36

### DIFF
--- a/src/components/AvatarRecordV2/index.tsx
+++ b/src/components/AvatarRecordV2/index.tsx
@@ -1,0 +1,129 @@
+import { generateAvatarInitials } from '@tetherto/pear-apps-utils-avatar-initials'
+import { rawTokens, useTheme } from '@tetherto/pearpass-lib-ui-kit'
+import { StarFilled } from '@tetherto/pearpass-lib-ui-kit/icons'
+import { useFavicon } from '@tetherto/pearpass-lib-vault'
+import { Image, StyleSheet, Text, View } from 'react-native'
+
+import { RECORD_COLOR_BY_TYPE } from '../../constants/recordColorByType'
+
+type AvatarSize = 'sm' | 'md' | 'lg'
+
+type RecordLike = {
+  type: string
+  data?: { title?: string }
+  isFavorite?: boolean
+}
+
+type Props = {
+  record: RecordLike
+  size?: AvatarSize
+  websiteDomain?: string
+  isFavorite?: boolean
+  testID?: string
+}
+
+const SIZES: Record<AvatarSize, number> = {
+  sm: 24,
+  md: 32,
+  lg: 40
+}
+
+const FONT_SIZES: Record<AvatarSize, number> = {
+  sm: rawTokens.fontSize12,
+  md: rawTokens.fontSize14,
+  lg: rawTokens.fontSize16
+}
+
+const STAR_SIZES: Record<AvatarSize, number> = {
+  sm: 10,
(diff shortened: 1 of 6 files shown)


boris91 (dismissed):

torkinos:
@nickgr2 please review

nickgr2 (changes requested):
A correctness issue in the new records query — and a minor cleanup.

**Bug: `isFavorite: false` excludes favorited records from the preview**

`ImportPreviewStep.tsx`:
```ts
const { data: records } = useRecords({
  shouldSkip: false,
  variables: {
    filters: {
      searchPattern: '',
      type: '',
      folder: '',
      isFavorite: false   // ← filters OUT favorites
    },
    ...
  }
})
```

`selectRecords` treats `isFavorite` as a strict equality filter when it's a boolean (`node_modules/@tetherto/pearpass-lib-vault/src/selectors/selectRecords.js`):

```js
if (
  typeof filters?.isFavorite === 'boolean' &&
  !!record.isFavorite !== filters.isFavorite
) {
  return false
}
```

So passing `false` drops every favorite from the result. A user importing a vault with any favorited records will see an under-count and missing rows in the preview — directly contradicting the "number of imported items" fix this PR is targeting.

The previous `getAllRecords(vault)` explicitly merged `vault.favorites?.records` into the list, so favorites used to be visible. Compare with `src/screens/Home/index.jsx`, which only adds the filter when the user actually toggles favorites: `...(state?.isFavorite ? { isFavorite: true } : {})`.

Suggested fix: drop the `isFavorite` key entirely (the same applies to `type`/`folder` — they're already harmless because the selector only activates them when truthy, but the consistent pattern elsewhere is to omit unused keys rather than send empty strings):

```ts
filters: {
  searchPattern: ''
}
```

**minor: `isPaired` is dead state in `useImportVault.ts`**

`pairWithCode` now returns a `boolean` and the caller (`index.tsx`) drives navigation off that. `state.isPaired` is set but no consumer reads it. Either wire it into the parent (e.g. drive `step` from it) or drop it and the state-shape change becomes a one-liner.

nikushaomanidze:
@nickgr2 Fixed your comments, please review again and approve if everything will be well

nickgr2 (approved):
Both prior comments addressed:

- `isFavorite: false` removed — favorites now appear in the import preview
- `isPaired` state dropped from `useImportVault`; `pairWithCode` cleanly returns `boolean` and the parent drives navigation off it

LGTM.

torkinos (approved):

boris91 (approved):

PR #341, opened 2026-04-29
V2 design fixes in creating items

denislupookov:
### Changes

[v2][Android/iOS][Creating Item] Bugs

### Screenshots/Recordings

https://github.com/user-attachments/assets/cca66fcc-4790-49eb-bd6f-24dd13c086aa

---
  - https://app.asana.com/0/0/1214338486310349

### DIFF
--- a/src/containers/BottomSheetFolderSelectorContent/index.jsx
+++ b/src/containers/BottomSheetFolderSelectorContent/index.jsx
@@ -12,7 +12,6 @@ import {
   CreateNewFolder,
   Folder,
   FolderCopy,
-  FolderOpen,
   MoreVert
 } from '@tetherto/pearpass-lib-ui-kit/icons'
 import { useFolders, useRecordCountsByType } from '@tetherto/pearpass-lib-vault'
@@ -26,8 +25,7 @@ import { Layout } from '../Layout'
 export const BottomSheetFolderSelectorContent = ({
   selectedFolder,
   onSelect,
-  includeAllFolders = true,
-  includeNoFolder = false
+  includeAllFolders = true
 }) => {
   const { t } = useLingui()
   const navigation = useNavigation()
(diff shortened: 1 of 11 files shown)


denislupookov:
@nickgr2 please review!

boris91 (dismissed):

nickgr2 (commented):
A few minor cleanups; nothing blocking.

**`renderHighlightedPassword` is now a no-op.** In `src/screens/CreateRecord/v2/CreatePasswordItemV2.tsx`, all three branches of the inner `map` callback return `part` unchanged, so the function just splits the password and re-joins it (with `null` entries from empty splits). The previous color-highlight branches were removed but the split/map scaffolding around them was left in place. Either drop the function and render `{generatedValue}` directly, or — if highlighting was intentionally dropped pending design — leave a comment explaining why. As-is the regex split serves no purpose.

minor: in `CreateOrEditPassphraseContent.tsx` the standalone Comment `InputField` now sits outside the `MultiSlotInput` but still uses `testID="comments-multi-slot-input-slot-0"`. The `multi-slot-input-slot` suffix no longer matches its position — consider renaming to e.g. `comment-input-field`.

minor: `CreatePasswordItemV2.tsx` picked up a stray double blank line just before the `<View style={[styles.groupedCard,…]}>` block.

Otherwise: the keyboard dismiss before navigating to the password generator, the `disableKeyboardAvoidance` opt-out for that screen, the `passwordIndicator` wiring, and the empty-title submit guards all read well. The `BottomSheetUploadFileContentV2` switch from internal `dismiss()` to `onOpenChange?.(false)` is fine — the only consumer (`AttachmentFieldsV2`) drives it in controlled mode.

denislupookov:
@nickgr2 addressed your changes, re review please

nickgr2 (approved):
Follow-up commit addresses all three earlier comments: `renderHighlightedPassword` now actually highlights digits via `theme.colors.colorPrimary` (and the regex no longer splits on non-alphanumerics), the standalone Comment input is renamed to `testID="comment-input-field"`, and the stray double blank line is gone. The `as never` cast on the inline `style` is regrettable but commented and contained — fine as-is.

Approving.

torkinos (approved):

boris91 (approved):

PR #342, opened 2026-04-29
[v2] [mobile] Locked screen

giorgikh93:
### Screenshots/Recordings

https://github.com/user-attachments/assets/2afbda81-953a-4e95-b77f-2261089f6c01

### dependencies

---
  - https://app.asana.com/0/0/1214386530275808

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7783,7 +7783,7 @@
     "node_modules/@tetherto/pearpass-lib-ui-kit": {
       "name": "pearpass-lib-ui-react-native-components",
       "version": "0.0.40",
-      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#135e27a5fbe5da6e56d0e210ada5233d8bc957f3",
+      "resolved": "git+ssh://git@github.com/tetherto/pearpass-lib-ui-react-native-components.git#2499e063482020abd394eadb9b878b8126f22d99",
       "license": "Apache-2.0",
       "dependencies": {
         "react-strict-dom": "^0.0.55"
--- a/src/containers/Auth/LockedScreenV2.jsx
+++ b/src/containers/Auth/LockedScreenV2.jsx
@@ -1,0 +1,187 @@
+import { useEffect } from 'react'
+
+import { useLingui } from '@lingui/react/macro'
+import { useNavigation } from '@react-navigation/native'
+import { useCountDown } from '@tetherto/pear-apps-lib-ui-react-hooks'
+import {
+  rawTokens,
+  Text,
+  PageHeader,
+  useTheme
+} from '@tetherto/pearpass-lib-ui-kit'
+import { WatchLater } from '@tetherto/pearpass-lib-ui-kit/icons'
+import { useUserData } from '@tetherto/pearpass-lib-vault'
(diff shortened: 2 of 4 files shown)


giorgikh93:
@nickgr2 please review

nickgr2 (changes requested):
A few issues from reading the diff:

**`src/containers/Auth/LockedScreenV2.test.jsx:423,512` — countdown test will fail**

The test asserts:

```jsx
expect(getByTestId('locked-screen-countdown').props.children).toBe('2:05')
```

…but `LockedScreenV2.jsx:121-126` renders the countdown without a `testID`:

```jsx
<Text
  variant="body"
  style={[styles.timerDigits, { color: theme.colors.colorPrimary }]}
>
  {timeRemaining}
</Text>
```

The kit `Text` mock here is `({ children, testID }) => <RN.Text testID={testID}>...`, so an undefined `testID` prop renders an undecorated node and `getByTestId('locked-screen-countdown')` will throw. Add `testID="locked-screen-countdown"` to that `Text` (REVIEW_GUIDELINES.md also asks for `testID` on E2E-relevant elements).

**`src/containers/Auth/LockedScreenV2.jsx:169` — hardcoded `borderRadius: 10`**

Everything else in this stylesheet uses `rawTokens.*`, but the timer strip drops to a literal:

```jsx
borderRadius: 10,
```

`rawTokens` (at the new ui-kit SHA) exposes `radius6`, `radius8`, `radius16`, `radius20`, `radius26` — pick the closest token (`radius8` looks right for a 1px-bordered strip), or add a `radius10` to the kit if the design genuinely calls for 10. Mixing one literal into an otherwise-tokenized stylesheet is the pattern we want to avoid.

**minor: `src/containers/Auth/LockedScreenV2.jsx:134` — `getStyles()` factory adds nothing**

`getStyles` doesn't close over `theme` or any prop, so it's invoked on every render only to re-call `StyleSheet.create`. Either hoist to a module-level `const styles = StyleSheet.create({...})` (as v1 does), or thread `theme` through if you intend to read theme tokens from it later. Right now it's the worst of both — extra work per render with no benefit.

**minor: typo in toast copy carried over from v1**

`text1: t\`Now you can enter you master password once again!\`` — "you" → "your". Carried verbatim from `LockedScreen.jsx:47`, so out of scope for this PR if you'd rather fix it separately, but the v2 rewrite is a natural moment to clean it up (and update the Lingui catalog once).

Everything else looks consistent with the kit-migration pattern: imports come from `@tetherto/pearpass-lib-ui-kit`, `V2` suffix is correct given the v1 sibling, `isV2()` routing is wired up at the call site, and `package-lock.json` bumps to a ui-kit SHA that does include the new tokens this file depends on (`spacing60`, etc.).

giorgikh93:
@nickgr2 addressed your points. Please review again

nickgr2 (approved):
All requested changes addressed at 3f4b1a3:

- Failing test fixed by switching to `getByText('2:05')` / `getByText('0:00')`
- `borderRadius: 10` now uses `rawTokens.spacing10`
- `getStyles()` factory hoisted to a module-level `styles` const
- `enter you master password` typo corrected

LGTM.

boris91 (approved):

hakaari (approved):

PR #344, opened 2026-04-30
Fix v2 record items and move flow bugs

denislupookov:
### Changes

[v2][Android/iOS][Create new Vault] Bugs
[v2][Android/iOS][Move Item flow] Bugs
[v2][Android/iOS][Item details content] Bugs

---
  - https://app.asana.com/0/0/1214338486310371
  - https://app.asana.com/0/0/1214338244933625
  - https://app.asana.com/0/0/1214338486310621

### DIFF
--- a/src/containers/Auth/VaultWizard/StepIdentity.jsx
+++ b/src/containers/Auth/VaultWizard/StepIdentity.jsx
@@ -16,12 +16,14 @@ import {
   getPasswordValidationMessages,
   getPasswordsMatch
 } from '../../../utils/passwordPolicy'
+import { unsupportedFeaturesEnabled } from '../../../utils/unsupportedFeatures'
 import { Layout } from '../../Layout'
 import { BackScreenHeader } from '../../ScreenHeader/BackScreenHeader'
 import { ConfirmablePasswordFields } from '../shared/ConfirmablePasswordFields'
 
 export const StepIdentity = ({ initialData, onSubmit, onBack }) => {
   const { t } = useLingui()
+  const isVaultPasswordEnabled = unsupportedFeaturesEnabled()
 
   const [name, setName] = useState(initialData?.name ?? '')
   const [usePassword, setUsePassword] = useState(
@@ -95,7 +97,11 @@ export const StepIdentity = ({ initialData, onSubmit, onBack }) => {
     >
       <PageHeader
         title={t`Create New Vault`}
-        subtitle={t`Create your vault by giving it a name. Add an optional vault password if you want an extra layer of protection on top of your master password.`}
+        subtitle={
+          isVaultPasswordEnabled
+            ? t`Create your vault by giving it a name. Add an optional vault password if you want an extra layer of protection on top of your master password.`
+            : t`Create your vault by giving it a name.`
+        }
         subtitleTestID="new-vault-subtitle"
       />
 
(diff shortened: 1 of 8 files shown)


denislupookov:
@nickgr2 please review!

nickgr2 (changes requested):
Reviewed against `CLAUDE.md`, `AGENTS.md`, and `REVIEW_GUIDELINES.md` at `3cf7d38`.

### Blocking

**`StepIdentity.test.jsx` will fail.** The second test (`requires matching strong passwords before submitting a protected vault`) calls `fireEvent.press(getByTestId('new-vault-password-toggle'))` at line 137, but the toggle is now wrapped in `{isVaultPasswordEnabled ? ... : null}`. `unsupportedFeaturesEnabled()` returns `UNSUPPORTED === true`, and `@tetherto/pearpass-lib-constants` exports `UNSUPPORTED = false`, so under Jest the toggle isn't rendered and `getByTestId` will throw. `src/screens/OnboardingV2/screens/CreatePasswordScreen.test.jsx` already shows the right pattern — `jest.mock('../../../utils/unsupportedFeatures', () => ({ unsupportedFeaturesEnabled: () => true }))`. Please mock the util in `StepIdentity.test.jsx` (and ideally add a third test that asserts the toggle is hidden when the flag is off).

**`MultiSelectDelete` scroll regression.** Removing `flex: 1` from both `recordsSection` and `recordsScroll` in `styles.js` breaks scrolling for long selections. `Layout`'s content view has `flex: 1`, but with no flex on the inner View or ScrollView the section sizes to its natural content height. `recordsLayoutHeight` (set in the `ScrollView` `onLayout`) will equal `recordsContentHeight`, so `showGradient = recordsContentHeight > recordsLayoutHeight` is always `false` and the ScrollView never scrolls. With many selected items the rows will overflow / push `confirmText` off-screen. `MultiSelectMove` works because its `recordsSection` gets a calculated `maxHeight: recordsMaxHeight` set inline; `MultiSelectDelete` has no equivalent. Either restore `flex: 1` on the section + ScrollView, or apply a similar `maxHeight` calculation.

### Questions

**`RecordDetailsV2` — Share button removed without replacement.** The diff drops the `Share` button (and `ShareIcon`, `RECORD_TYPES`, `Share` from `react-native`) from the header `rightActions`, but `BottomSheetRecordActionsContentV2` / `useRecordActionItems` don't surface a share action either, so the share entry point disappears entirely. PR description mentions only bug fixes for "Item details content" — was removing share intentional, and if so, where does it live now?

**`useGetMultipleFiles.js` — `updatedAt` dep.** Adding `initialRecord?.updatedAt` to the deps array is fine for refetching when the record changes, but the existing `JSON.stringify(initialRecord?.data?.[fieldName] || [])` already covers attachment-data changes. Was the intent to also refetch when non-attachment fields update (e.g., title, isFavorite, folder)? If yes, OK; if the goal was only to fix stale attachments after `updateFolder`, narrower deps would avoid the extra effect runs.

### Minor

- `PassPhraseV2.tsx`: the new `<Button aria-label="Copy recovery phrase" iconBefore={<CopyIcon ...>}>` uses `CopyIcon` from `@tetherto/pearpass-lib-ui-react-native-components` (legacy v1). The kit ships `ContentCopy` at `@tetherto/pearpass-lib-ui-kit/icons` — please switch new usages to the kit icon. The legacy import already exists in this file, but each new usage entrenches it further.
- `PassPhraseV2.tsx`: `aria-label="Copy recovery phrase"` is a hardcoded English string. Pass it through `t\`...\`` to match the project's i18n rule.
- `PassPhraseV2.tsx`: the new `<Text variant="label">{t\`Recovery Phrase\`}</Text>` and the new copy `<Button>` have no `testID`. The kit `Button` accepts `testID`/`data-testid` — please add them for E2E.
- `StepIdentity.jsx`: when `isVaultPasswordEnabled` is false, the unused `usePassword` / `password` / `passwordConfirm` state and the related `useMemo`s are still allocated. Not a bug — just noting that lifting them inside the conditional branch (or extracting a `<VaultPasswordFields>` subcomponent) would be cleaner if the flag is expected to stick around.

boris91 (dismissed):

denislupookov:
@nickgr2 addressed your requests, re review please!

1. Share button: intentional — good catch. The bottom sheet Share Item is gated behind UNSUPPORTED and has a no-op onPress, so removing the header button leaves no working entry point. I'll restore the header button (or wire up the bottom sheet one) before merging.

2. updatedAt dep: the goal was just to fix stale attachments after updateFolder, not to refetch on every field change. I'll narrow it to initialRecord?.folder so we don't redundantly re-fetch files on title/favorite edits.

nickgr2 (changes requested):
Re-reviewed at `9539f68` against the diff since `3cf7d38`.

### Resolved
- `StepIdentity.test.jsx`: module-level `jest.mock('../../../utils/unsupportedFeatures', ...)` with default `jest.fn(() => true)` plus a third test asserting the toggle hides when the flag is off. Matches the pattern.
- `MultiSelectDelete`: wrapper container with `flex: 1`, `recordsMaxHeight = containerHeight - confirmTextHeight` applied as `maxHeight` on `recordsSection`. Same shape as `MultiSelectMove`'s proven pattern; scrolling will work for long selections.
- `PassPhraseV2.tsx`: kit icons (`ContentCopy`/`ContentPaste`) replace the legacy ones, `aria-label` is wrapped in `` t`` ``, and `data-testid` is added on the label and copy button.

### Still required
- **Share entry point still missing on V2 details.** Per your reply, the header `Share` removal was unintentional and the bottom-sheet share is `UNSUPPORTED`-gated with a no-op. Verified at head: `src/screens/RecordDetails/v2/RecordDetailsV2.jsx` has no `Share` reference. Either restore the header button or wire up the bottom-sheet entry.
- **`useGetMultipleFiles.js` deps.** Per your reply, the intent is to narrow to `initialRecord?.folder`. Line 168 still references `initialRecord?.updatedAt`.

### Minor
- `PassPhraseV2.tsx` lines 9–10: the named-import block from `@tetherto/pearpass-lib-ui-react-native-components` is now empty (`import { } from '...'`). Drop the import entirely.

denislupookov:
@nickgr2 the required changes were intentional:

we do not need Share anymore in the header.
updatedAt is the right dep because saving edits to a media item bumps updatedAt but doesn't reliably change the field JSON.

denislupookov:
@nickgr2 re review please

nickgr2 (approved):
All previously-flagged items resolved — both outstanding concerns from the prior review (Share entry point on V2 details, `useGetMultipleFiles` dep on `updatedAt`) were addressed via the explanations in the comment thread, which I've accepted.

Approving to clear the prior `REQUEST_CHANGES`.

PR #353, opened 2026-04-30
Fix settings / item details / home bugs

denislupookov:
### Changes

[v2][Android/iOS]["Create master password" screen] Update password hint
[v2][Android/iOS][Settings][Blind Peering] Bugs
[v2][Android/iOS][Settings][Report a Problem] Bugs

---
  - https://app.asana.com/0/0/1214338486310321
  - https://app.asana.com/0/0/1214387227184036
  - https://app.asana.com/0/0/1214404999323681

### DIFF
--- a/src/components/FolderSelectField/index.tsx
+++ b/src/components/FolderSelectField/index.tsx
@@ -1,0 +1,60 @@
+import { useLingui } from '@lingui/react/macro'
+import {
+  ContextMenu,
+  MultiSlotInput,
+  SelectField,
+  useTheme
+} from '@tetherto/pearpass-lib-ui-kit'
+import { KeyboardArrowBottom } from '@tetherto/pearpass-lib-ui-kit/icons'
+
+import { BottomSheetFolderSelectorContent } from '../../containers/BottomSheetFolderSelectorContent'
+
+type Props = {
+  value?: string
+  onChange: (folderName: string) => void
+  testID?: string
+  multiSlotTestID?: string
+}
+
+export const FolderSelectField = ({
+  value,
+  onChange,
+  testID = 'folder-field',
+  multiSlotTestID = 'folder-multi-slot-input'
+}: Props) => {
+  const { t } = useLingui()
+  const { theme } = useTheme()
+
+  const handleSelect = (folder?: { name?: string }) => {
+    if (!folder) return
+    if (folder.name === 'allFolder') {
+      onChange('')
+      return
+    }
+    onChange(folder.name === value ? '' : (folder.name ?? ''))
+  }
+
+  return (
+    <MultiSlotInput testID={multiSlotTestID}>
+      <ContextMenu
+        trigger={
+          <SelectField
+            label={t`Folder`}
+            value={value ?? ''}
+            placeholder={t`Choose Folder`}
+            isGrouped
+            testID={testID}
+            rightSlot={
+              <KeyboardArrowBottom color={theme.colors.colorTextPrimary} />
+            }
+          />
+        }
+      >
+        <BottomSheetFolderSelectorContent
+          selectedFolder={value}
+          onSelect={handleSelect}
+        />
+      </ContextMenu>
+    </MultiSlotInput>
+  )
+}
--- a/src/containers/BottomSheet/UnsavedChangesSheet.jsx
+++ b/src/containers/BottomSheet/UnsavedChangesSheet.jsx
@@ -1,0 +1,57 @@
+import { useLingui } from '@lingui/react/macro'
+import { Button, Text, rawTokens } from '@tetherto/pearpass-lib-ui-kit'
+import { StyleSheet, View } from 'react-native'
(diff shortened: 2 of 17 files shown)


denislupookov:
@nickgr2 please review

boris91 (dismissed):

nickgr2 (changes requested):
Reviewed `97fa9af`. One correctness concern in the unsaved-changes flow worth fixing before merge — rest are minor/style.

### `BlindPeeringSectionV2.jsx` — stale closure in `useFocusEffect` listener

```js
useFocusEffect(
  useCallback(() => {
    const sub = navigation.addListener('beforeRemove', (e) => {
      ...
      onSave={async () => {
        collapse()
        const ok = await performSave()
        ...
      }}
    })
    return sub
  }, [isDirty, navigation, expand, collapse])
)
```

`performSave` (and `peerMode`, `blindPeersList`, `isEnabled` it closes over) is captured at the render where `useCallback` was last invoked — i.e. the render where `isDirty` flipped `false → true`. After that, while `isDirty` stays `true`, deps don't change, so `useFocusEffect` keeps the same listener with the same closure.

Concrete scenario: user types one char in peer field (isDirty becomes true → listener re-binds), then types more chars (isDirty stays true → listener NOT re-bound, still closes over the first-char state). User presses back, taps "Save Changes" in the sheet → `performSave` runs against stale `blindPeersList`, persisting only the first char and dropping subsequent edits. Same risk for `peerMode` toggles that don't flip dirtiness.

Suggested fix: keep a ref of `performSave` (or of the live state) and read `.current` inside the listener; or add the underlying state to the `useCallback` deps so the listener re-binds on any change.

### `BottomSheetVaultSelectorContent/index.jsx` & `Vaults2/VaultRow.jsx` — pluralization (minor)

```js
itemCount === 1 ? t`${itemCount} Item` : t`${itemCount} Items`
deviceCount === 1 ? t`${deviceCount} Device` : t`${deviceCount} Devices`
memberCount === 1 ? t`${memberCount} Member` : t`${memberCount} Members`
```

Consider Lingui's `plural` macro — `plural(itemCount, { one: '# Item', other: '# Items' })` — so locales with more than two plural forms (Russian, Polish, Arabic, etc.) work correctly.

### `CreateOrEditLoginContent.tsx` — not migrated (minor)

All other `CreateOrEdit*Content.tsx` v2 screens were swapped to the new `FolderSelectField`. `CreateOrEditLoginContent.tsx` still has the inline `ContextMenu`/`SelectField`/`BottomSheetFolderSelectorContent` block (with `handleFolderSelect` whose body is identical to `FolderSelectField`'s handler). Worth migrating in this PR or a follow-up to keep the Folder field consistent across record types.

### Tests (minor)

- `VaultRow.test.jsx` exists but isn't extended for the new `{primary, secondary}` subtitle (existing cases pass `vault` without `records`/`devices`, so they still pass — but the new branch is uncovered).
- No tests added for `FolderSelectField`, `UnsavedChangesSheet`, or the `isDirty` / `beforeRemove` flow.

### Things checked OK

- `BLIND_PEERS_LEARN_MORE` is exported from `@tetherto/pearpass-lib-constants` (verified in node_modules).
- `Link` on native: kit's `useLinkPress()` is called without `href` and returns undefined, so the explicit `onClick={() => Linking.openURL(...)}` workaround is necessary.
- `disabled={loadingAction !== null}` removal is fine — `launchPicker` resets `loadingAction` on the next animation frame, so `isLoading` already reflects the actual async state.
- `Layout` Android `behavior: undefined` — matches RN guidance (let `windowSoftInputMode="adjustResize"` handle it natively).
- `Radio` `builtIn` prop is real; `ListItem` `subtitle` accepts `string | { primary, secondary }` per kit types.

denislupookov:
@nickgr2 please re review

Note: We do not use plural this time, let's pass it

nickgr2 (approved):
Re-reviewed `40ad0a7`. Both items from my prior review are addressed correctly.

### `BlindPeeringSectionV2.jsx` — stale closure fixed

```js
const performSaveRef = useRef(performSave)
useEffect(() => {
  performSaveRef.current = performSave
})
```

Standard latest-ref pattern: dep-less `useEffect` updates the ref on every render, and the `beforeRemove` listener reads `performSaveRef.current()` so it always invokes the freshest closure (with the live `peerMode` / `blindPeersList` / `isEnabled`). Resolves the original concern.

### `CreateOrEditLoginContent.tsx` — migrated to `FolderSelectField`

Inline `ContextMenu`/`SelectField`/`BottomSheetFolderSelectorContent` block and `handleFolderSelect` removed; replaced with `<FolderSelectField value={values.folder} onChange={(val) => setValue('folder', val)} />`. Imports cleaned up (`ContextMenu`, `SelectField`, `KeyboardArrowBottom`, `BottomSheetFolderSelectorContent` all dropped). `FolderSelectField` encapsulates identical select/toggle logic, so behavior is preserved and the Folder field is now consistent across record types.

torkinos (approved):

PR #355, opened 2026-05-01
Fixes for settings v2 bugs

denislupookov:
### Changes

[v2][Android/iOS][Settings][Import items] Bugs
[v2][Android/iOS][Settings][App version/ Language screen] Bugs
[v2][Android/iOS][Settings][Master Password] Bugs

---
  - https://app.asana.com/0/0/1214338486310686
  - https://app.asana.com/0/0/1214338486310309
  - https://app.asana.com/0/0/1214387152114112

### DIFF
--- a/src/screens/ImportItems/index.tsx
+++ b/src/screens/ImportItems/index.tsx
@@ -629,7 +629,7 @@ export const ImportItems = () => {
                     onClick={() => Linking.openURL(selectedOption.supportLink!)}
                     data-testid="import-support-link"
                   >
-                    {t`learn more about exporting data from ${selectedOption.title}`}
+                    {t`Learn more about exporting data from ${selectedOption.title}`}
                   </Link>
                   .
                 </>
--- a/src/screens/ImportItems/index.tsx
+++ b/src/screens/ImportItems/index.tsx
@@ -629,7 +629,7 @@ export const ImportItems = () => {
                     onClick={() => Linking.openURL(selectedOption.supportLink!)}
                     data-testid="import-support-link"
                   >
-                    {t`Learn more about exporting data from ${selectedOption.title}`}
+                    {t`learn more about exporting data from ${selectedOption.title}`}
                   </Link>
                   .
                 </>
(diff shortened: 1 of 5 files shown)


denislupookov:
@nickgr2 review please

boris91 (approved):

nickgr2 (approved):
Reviewed against `REVIEW_GUIDELINES.md` (kit compliance, styling, RN correctness).

**Changes look correct.** All four edits target real issues and use kit components properly:

- `LoginRecordDetailsForm.tsx` — moving `websites` out of form values is consistent with this being a read-only details form. The other read-only fields use `register(...)` because they're plain strings; arrays don't round-trip through the form helper as cleanly, so reading directly from `initialRecord?.data?.websites` is reasonable. `useMemo` deps are still correct since `initialRecord` is the only source.
- `AboutV2.jsx` — good catch removing the `<Text as="span" variant="label">` wrapper around the subtitle. `PageHeader` already wraps `subtitle` in `<Text as="p" variant="label">` (kit `PageHeader.tsx:23`), so the prior code was nesting Text-in-Text. Switching to `<>` is the correct fix.
- `AppearanceV2.jsx` — replacing the manual `NavbarListItem.map(...)` with the kit's `Radio` component is the right move; the prior pattern was effectively rolling a radio group out of list items. New padding values use `rawTokens` — kit-compliant.
- `MasterPassword/index.jsx` — `getRepeatIndicator` returning `undefined` instead of `'vulnerable'` is a sensible fix since the indicator semantically reflects strength, not match state. The submit button stays disabled while `values.newPassword !== values.repeatPassword` (line 85), so the user still gets feedback that the form isn't ready.

**Minor (non-blocking):**
- `AppearanceV2.jsx` `LanguagePickerSheet`: the new `Radio` doesn't get a `testID`, so E2E can't target language options inside the sheet. The old `NavbarListItem.map` didn't have per-option `testID`s either, so no regression — but a `testID="language-radio"` on the `Radio` would be a cheap follow-up.
- `MasterPassword`: while typing the repeat password, the user gets no positive cue that it doesn't match other than the disabled button. Not a blocker for this fix, but a future "mismatch" indicator state in the kit would be the cleaner UX.

PR #396, opened 2026-05-08
fdroid related changes

ndrkltsk:
added script for building bare from sources, required by fdroid.
added new metadata for fdroid.


ndrkltsk:

boris91 (dismissed):

nickgr2 (approved):
A few non-blocking observations on the build script and metadata changes.

**Wiring — confirm intent**

`scripts/fdroid/build-bare-from-source.sh` isn't invoked by `scripts/fdroid-build.sh` or any `npm run` script in the repo. Presumably the F-Droid recipe in `fdroiddata` calls it directly between `npm ci` and gradle's `preBuild` (the script reads from `node_modules/bare-link` and writes prebuilds back into `node_modules/<addon>/prebuilds/` for `link.mjs` to pick up). Confirming this is intentional and won't get orphaned.

**`minor:` prereq check is incomplete**

`discover_addons` reads `node_modules/bare-link/lib/dependencies` (script line 379), but `check_prereqs` only validates `react-native-bare-kit`. `bare-link` is currently hoisted as a transitive of `react-native-bare-kit@0.13.0` — if a future kit upgrade changes its dep tree, the script would fail with `MODULE_NOT_FOUND` from Node rather than the cleaner `fatal` message. A one-liner near the other directory checks — `[ -d "$PROJECT_DIR/node_modules/bare-link" ] || fatal "bare-link missing; run 'npm ci' first"` — would keep diagnostics consistent.

**`minor:` changelog covers only two ABIs**

`metadata/en-US/changelogs/16101.txt` and `16102.txt` match the per-ABI versionCode scheme from `scripts/fdroid-build.sh` (`versionCode * 10 + abiOffset`, armeabi-v7a=1, arm64-v8a=2). x86 and x86_64 (offsets 3, 4) get no changelog. If the F-Droid recipe ships only the two arm ABIs that's fine; otherwise emulator users would see an empty release-notes panel.

**`minor:` dropped F-Droid build notes from full description**

The previous `full_description.txt` ended with a paragraph noting that the Android "fdroid" channel removes Play Services–specific auth dependencies and disables in-app version checking / Play Store redirects. That's useful context for users comparing this build to the Play Store version (different feature set on the same app page). Worth keeping it (or compressing to a "What's different on F-Droid" line) rather than removing outright.

ndrkltsk:
@nickgr2 

### Wiring
confirmed intentional.
The fdroid recipe in fdroiddata/metadata/com.pears.pass.yml calls the script directly from the build phase:                                                      
                                                                                                                                                                                             
```
  build:                                                                                                                                                                                     
    - bash scripts/fdroid/build-bare-from-source.sh arm64-v8a                                                                                                                                
    - cd android
    - gradlew-fdroid assembleFdroidRelease --no-daemon --parallel
```

###  prereq check
Good catch, fixed

### covers only two ABIs 
Intentional. 
The fdroid recipe ships only armeabi-v7a (16101) and arm64-v8a (16102). x86/x86_64 (offsets 3, 4) aren't in the recipe, so no changelogs needed for them. If we add emulator support later we'll add then.

### dropped F-Droid build notes from full description
Acknowledged, but keeping the description as-is for now. 
The differences are real, but we're treating them as the fdroid baseline rather than diffs the user needs to reason about. Open to revisiting if other reviewers feel strongly.

torkinos (approved):

boris91 (approved):

PR #400, opened 2026-05-11
Remove logic

giorgikh93:
### Changes

- OTP Delete approach without removing Login record 
### Testing Notes

- Tested locally on ios simulator
### Screenshots/Recordings

https://github.com/user-attachments/assets/ff48a564-fa6a-4614-bcb1-0227cc2475e1


giorgikh93:
@nickgr2 please review

nickgr2 (commented):
Reviewed at 17abea65. The OTP-context plumbing (`recordType: OTP` from Authenticator → `isOtpContext` through RecordDetailsV2 → MultiSelectDelete) is coherent, and stripping `data.otp` / `data.otpInput` is the right surgical update (vault re-derives `otpPublic` on next listRecords). A few items worth addressing:

**Inconsistent copy in MultiSelectDelete when stripping OTP** — `src/screens/MultiSelectDelete/index.jsx`
Only the header swaps to "Remove Authenticator Code". The button still reads `Delete Item` and the confirm caption still says `Are you sure to delete the selected item?`. For the `isSingleLoginInOtpContext` branch these should match the "remove" framing (e.g. `Remove`, `Are you sure you want to remove the authenticator code from this login?`), otherwise users see a delete confirmation for a non-delete action.

**`delete updatedRecord.otpPublic` is dead code** — `src/screens/MultiSelectDelete/index.jsx:55-56`
`updateRecordsFactory` (in `@tetherto/pearpass-lib-vault`) only copies `id, type, vaultId, data, folder, isFavorite, createdAt, updatedAt`. `otpPublic` never reaches the API, so the local delete has no effect. Either drop it, or add a one-line comment if it's there for clarity.

**Test coverage for the new branch** — no spec exercises `isOtpContext` / `isAuthenticatorLoginRecord`. `useRecordActionItems.test.js` already covers delete-action presence; a case that asserts `handleDelete` navigates to `MultiSelectDelete` with `isOtpContext: true` when `recordType=OTP` + `record.type=LOGIN`, and that `handleEdit` navigates with `recordType: OTP` in that case, would lock in the new behavior cheaply. The strip-OTP handler in `MultiSelectDelete` has no test file at all — the file is uncovered today, but this PR is the first to add non-trivial logic there.

**v1 parity (informational, not blocking)** — the v1 branch of `useRecordActionItems.handleDelete` still calls `deleteRecords([record?.id])` directly. With `MOBILE_DESIGN_VERSION = 2` this is effectively unreachable, but worth a follow-up note if v1 is still expected to work for any tester.

**Minor** — the placeholder change `Enter Secret Key (TOTP)` → `Enter your key or URI` reads slightly less precise. Consider `Enter secret key or otpauth:// URI` if URI input is supported.

giorgikh93:
@nickgr2 I added comment for clarity. Regarding delete texts, we're okay to have generic texts. Please review

nickgr2:
> Regarding delete texts, we're okay to have generic texts.

Acknowledged — generic copy is a reasonable call.

The new clarifying comment on `delete updatedRecord.otpPublic` resolves point 2.

Still open from the prior review:
- Test coverage for the new `isOtpContext` / `isAuthenticatorLoginRecord` branch (`useRecordActionItems` navigation + the strip-OTP path in `MultiSelectDelete`).
- Placeholder copy nit (`Enter your key or URI`) — minor.

The v1-parity note was informational. No re-review needed for the comment-only change.

giorgikh93:
@nickgr2 added tests. Enter your key or URI text is intentional per product requirements. Please review

nickgr2:
> added tests. Enter your key or URI text is intentional per product requirements

Tests look thorough — both the `useRecordActionItems` OTP routing and the `MultiSelectDelete` strip-OTP path are covered. Placeholder rationale is fine, that's a product call.

One item from the prior review isn't addressed though: in `src/screens/MultiSelectDelete/index.jsx` the header swaps to "Remove Authenticator Code" for `isSingleLoginInOtpContext`, but the button still uses "Delete Item" (line 96) and the confirm caption still says "Are you sure to delete the selected item?" (line 157). Was that intentional, or just missed? If intentional, fine — otherwise the OTP branch should use matching "Remove" copy throughout.

I'll do a fresh re-review pass shortly.

giorgikh93:
@nickgr2 Delete Item and "Are you sure to delete the selected item?" are intentional, we can proceed

nickgr2 (commented):
Re-reviewed at 5ee0b194. Incremental changes since 17abea65 are the two new test files plus the clarifying comment on `delete updatedRecord.otpPublic`.

**Tests** — coverage matches the prior review ask. `useRecordActionItems.test.js` exercises both branches of `isAuthenticatorLoginRecord` for `handleDelete` (asserts `isOtpContext: true`/`false` on the navigate payload) and `handleEdit` (asserts `recordType: OTP` vs `record.type`). The new `MultiSelectDelete/index.test.jsx` covers strip-OTP success (asserts `updateRecords` called with `otpInput`/`otp`/`otpPublic` stripped, `deleteRecords` not called), the error path (toast shown, `goBack` not called), and the fall-through delete path when `isOtpContext: false`. Mocks are at the right level.

Minor: the `useRecords` mock returns `{ deleteRecords, updateRecords }` directly and never invokes the `onCompleted` callback passed by the screen, so the success-path `navigation.goBack()` isn't asserted. Not blocking — the strip behavior is what matters.

**Still open from prior review** — the OTP-strip copy in `MultiSelectDelete/index.jsx` is half-converted. The header swaps to `Remove Authenticator Code` for `isSingleLoginInOtpContext`, but the button label at line 96 still reads `Delete Item` and the confirm caption at line 157 still says `Are you sure to delete the selected item?`. Users in the strip path see a delete confirmation for a non-delete action. Suggest gating the button label and the caption on `isSingleLoginInOtpContext` the same way the header is — e.g. button `Remove` and caption `Are you sure you want to remove the authenticator code from this login?`.

giorgikh93:
@nickgr2 updated delete texts, please reveiw

nickgr2 (commented):
Copy is now consistent across the title / button / caption for the single-login OTP-strip case — addresses my earlier flag.

No new findings on this delta. Test coverage for the `isSingleLoginInOtpContext` branch from my earlier review is still open but not a blocker for this copy fix.

giorgikh93:
@nickgr2 updated tests, please review

boris91 (approved):

nickgr2 (commented):
Test coverage for the isSingleLoginInOtpContext branch lands cleanly. The OTP path asserts header (Remove Authenticator Code), button (Remove), and caption copy, and the inverse case asserts the fallback Delete Item/caption — addressing the test-coverage gap flagged earlier. No new findings on this delta.

torkinos (approved):

PR #402, opened 2026-05-13
chore: update package-lock.json and remove obsolete GitHub workflows

torkinos:


torkinos:
@nickgr2 please review

kapildev421 (approved):

nickgr2 (approved):
LGTM as a cleanup PR.

- `build-android-fdroid.yml` had `if: false # Temporarily disabled`, so removing it has no behavioral impact.
- `scripts/fdroid-build.sh`: dropping `--install-links` from `npm ci` is a no-op here — `package.json` has no `file:` deps (all `@tetherto/*` workspace packages are `git+https://`), so the flag never affected resolution.
- `package-lock.json` diff is routine transitive bumps and dedup of nested duplicates (`react-is`, `isarray`); no new top-level packages.

One thing worth confirming before merge: `build-and-publish.yml` was the only workflow producing the EAS builds, APKs/AABs/IPAs, and GitHub releases on pushes to `main`/`dev`/`staging`. `eas.json` is preserved, so EAS itself still works, but if there's no replacement (separate pipeline, manual `eas build`, etc.) then automatic store submissions and tagged releases will stop after merge. If that replacement already exists elsewhere, ignore this — just flagging it since "obsolete" in the title implies one does.

hakaari (approved):

PR #405, opened 2026-05-14
Wire vault deletion UI and connect receive-side to action-exchange.

shavtvalishvili:
## Summary
Mobile delete-vault UX wired end-to-end with the receive-side action handler and the new direct-delivery transport (merged in from `device-direct-actions`).

### UI
- `VaultDeleteScreenV2` is a stack screen mirroring `VaultRenameScreen`: back-arrow header, master-password confirm, "Erase Vault from N other devices" toggle. Replaces the V2 confirmation modal.
- `PairedDevicesScreen` ports the inline device list into a stack screen so "View Paired Devices" matches Rename/Delete navigation-wise. `BottomSheetPairedDevicesContent` is retained for the in-screen "N other devices" Link peek inside the delete screen.
- `BottomSheetVaultAction` gains "View Paired Devices" (switches the active vault, then pushes the screen) and uses the standard `ArrowBackOutined` back icon.
- Paired-device count pluralised (`1 other device` vs `N other devices`).

### Receive-side
- `useVaultAccessRevoked` subscribes to `pearpassVaultClient.on('vault-access-revoked', ...)`, wipes the local vault, and creates a fresh "Personal" vault when nothing remains.
- Bails out when the vault is no longer in the local list (covers duplicate/self deliveries).

### Send-side
- "Erase from all devices" toggle calls `broadcastDeleteVault(vaultId)` before `deleteVaultLocal`. Per-target failure surfaces a toast; local deletion still proceeds.

### device-direct-actions (now merged)
- Cross-device delivery now goes through the personal-swarm direct transport with outbox-backed retries (1m for 7d, 1h to 30d, drop). Inbound envelopes are ed25519-signed using the sender's master-vault writer keypair and verified before being placed on the inbox. Inbox drops envelopes whose `actor` matches our own deviceId. See lib-vault and lib-vault-core PRs for the protocol.

## Recordings

https://github.com/user-attachments/assets/5bf0896e-0b49-4e2f-9aa6-4888d36fb3e9

https://github.com/user-attachments/assets/60ae5122-6ade-4b2e-97e0-c2684af23eb5

## Depends on
- https://github.com/tetherto/pearpass-lib-vault/pull/48
- https://github.com/tetherto/pearpass-lib-vault-core/pull/59


torkinos:
@nickgr2 please review

nickgr2 (commented):
A few items to consider; nothing blocking. UI kit usage, theming, naming, and i18n all look good.

**Missing tests for new behaviour**
- `VaultDeleteScreenV2/index.jsx`: master-password verify → optional `broadcastDeleteVault` → `deleteVaultLocal` → fallback "Personal" vault creation is a destructive multi-step flow. The v1 sibling `VaultDeleteScreen/` has `index.test.jsx`; the V2 does not.
- `hooks/useVaultAccessRevoked.jsx`: wipes a vault on a remote signal, creates a replacement, opens a modal. Worth a test — at minimum that it bails when `vault` isn't in the list, and that it falls back to creating a Personal vault when the active vault is wiped.
- `screens/Settings/PairedDevicesScreen/index.jsx` and `containers/Modal/AccessRemovedModalContent/index.jsx` are simpler but new screen/modal additions in this repo generally land with a test.

**Duplicated helpers**
`PairedDevicesScreen/index.jsx` reimplements `formatDeviceDate` and `getDeviceIcon` with the exact same bodies already in `containers/BottomSheetPairedDevicesContent/index.jsx`. Extract to e.g. `src/utils/devicePresentation.js` so the two views can't drift.

**`formatDeviceDate` is hardcoded to `en-US`**
`date.toLocaleString('en-US', { month: 'short' })` — the rest of the app is wired through Lingui. Either pass the active locale or drop the locale arg and let the platform default apply. (Same in the existing sheet — but worth fixing in both while you're here.)

**"Last used on" is the same value as "Paired on"**
In `PairedDevicesScreen` both `Last used on ${formatDeviceDate(device.createdAt)}` and `Paired on ${formatDeviceDate(device.createdAt)}` read `device.createdAt`. The first label is misleading until the underlying field exists — consider hiding the row entirely until there's a real `lastUsedAt`, rather than guarding it behind `unsupportedFeaturesEnabled()`.

**`VaultDeleteScreenV2` — redundant validation**
The form is set up with a Validator schema that requires `masterPassword`, and `handleSubmit(onSubmit)` already enforces it. The explicit `if (!formValues.masterPassword) { setErrors(...); return }` inside `onSubmit` is unreachable. Either remove it or remove the schema.

**`BottomSheetVaultAction` now always offers "Delete Vault"**
Previously the Delete entry was nested inside the `personalVault && !PROTECTED_VAULT_ENABLED` branch; in the new layout it's appended unconditionally. Confirm this is intentional — protected and non-personal vaults now show the option too. If it is intentional, the personal-only branch can be simplified (it's now just "Share Personal Vault").

**Minor: stable key in PairedDevicesScreen**
`key={device.name + index}` is fine when the list is stable but `device.id` would be the conventional choice for a list of paired devices (matches what `useVaultAccessRevoked` keys off too).

shavtvalishvili:
@nickgr2, I addressed:
- Tests added: `useVaultAccessRevoked`, `VaultDeleteScreenV2`, `AccessRemovedModalContent`.
- `formatDeviceDate` and `getDeviceIcon` extracted to `src/utils/devicePresentation.js`; both consumers updated.
- `formatDeviceDate` no longer hardcodes `'en-US'`.
- "Last used on" row removed from both `PairedDevicesScreen` and `BottomSheetPairedDevicesContent`.
- `VaultDeleteScreenV2` redundant `masterPassword` check removed.
- `PairedDevicesScreen` key switched to `device.id ?? \`${device.name}-${index}\``.
- Toast on fallback Personal-vault failure.
- Comment on `logIn` reuse for password verification.

I didn't address:
- `BottomSheetVaultAction` unconditional Delete is intentional in the V2 design.
- `PairedDevicesScreen` test left out; the screen is presentation over `useVault().data.devices` with no logic beyond the shared `devicePresentation` helpers.

giorgikh93 (approved):

nickgr2 (commented):
Re-reviewing dcb05c4 against e9fddcf. Most prior polish findings are addressed — extracted `devicePresentation`, dropped hardcoded `en-US`, removed the redundant "Last used on" `createdAt` line, switched device keys to `device.id`, added tests for `useVaultAccessRevoked` and `VaultDeleteScreenV2`. The two declined items (unconditional Delete in `BottomSheetVaultAction`, presentation-only `PairedDevicesScreen` skipping a test) are reasonable.

Two new issues from this revision:

**1. `VaultDeleteScreenV2` reads the active vault, not the deletion target**
`src/screens/Settings/VaultDeleteScreenV2/index.jsx:30` does `const { data: vaultData, ... } = useVault()`, but the screen is opened with `route.params.vaultId` from `BottomSheetVaultSelectorContent.onDelete` (`src/containers/BottomSheetVaultSelectorContent/index.jsx:115`) and `Vaults2.onDelete` (`src/screens/Settings/Vaults2/index.jsx:66`) without switching active vault first. When deleting a non-active vault, `vaultData.devices` is the wrong vault's devices, so `deviceCount` and `showEraseFromAllDevices` reflect the wrong vault — the "Erase Vault from all N devices" toggle/copy will be wrong (and may be hidden when it should appear, or vice versa).

The test only exercises the same-vault case (active=v1, target=v1), so it doesn't catch this.

Suggested fix: scope to the target vault — `useVault({ variables: { vaultId } })` (the type already supports this in `src/types/pearpass-lib-vault.d.ts:121-123`).

**2. `PairedDevicesScreen` has the same scoping issue**
`src/screens/Settings/Vaults2/index.jsx:57` does `onViewPairedDevices: () => navigation.navigate('PairedDevicesScreen')` without passing a vaultId and without switching first. `PairedDevicesScreen` uses bare `useVault()`. Clicking "View Paired Devices" on a non-active vault row shows the active vault's devices.

Note the selector path (`BottomSheetVaultSelectorContent` L111-113) handles this by `await switchVault(vault)` before navigating, so the two callers behave differently. Either match the selector pattern in `Vaults2`, or pass `vaultId` through and use `useVault({ variables: { vaultId } })` in `PairedDevicesScreen`.

**Minor**
- `PairedDevicesScreen` has no empty state when `devices.length === 0` — renders the header with an empty body. A short "No devices paired" message would be friendlier.
- `formatDeviceDate` now accepts a `locale` arg but `PairedDevicesScreen` and `BottomSheetPairedDevicesContent` both call it without one — which is fine for now (system locale), just flagging since the locale knob is unused.

PR #406, opened 2026-05-14
feat: bitwarden encrypted import

ndrkltsk:
Added also `--defer @tetherto/swarmconf` to fix bare worklet building

---
  - https://app.asana.com/0/0/1213323394838666


nickgr2:
On it.

nickgr2 (changes requested):
Two issues from this diff that should be addressed before merge.

**1. Regression: Bitwarden CSV imports will fail.** `src/screens/ImportItems/index.tsx` lines 285-288:

```ts
if (
  resolvedType === ImportOptionType.Bitwarden &&
  JSON.parse(fileContent as string).encrypted
) {
```

Bitwarden's `accepts` is `['.json', '.csv']` (line 94). For a CSV upload, `fileContent` is CSV text, `JSON.parse` throws `SyntaxError`, which is then swallowed by the outer `catch (err)` and re-thrown as `"Failed to decrypt file. Please check your password and try again."` — a misleading error for a previously working path. Gate on file type first, e.g.:

```ts
if (
  resolvedType === ImportOptionType.Bitwarden &&
  fileType === 'json' &&
  JSON.parse(fileContent as string).encrypted
) {
```

(or parse once into a local, reused below.)

**2. Security: `console.log` of decryption error in production.** Line 306:

```ts
} catch (err) {
  console.log('[bw-import] decrypt failed:', err)
```

Per the project's security guidelines, production code should not log error bodies that may surface user-derived crypto state. This looks like leftover debug — please remove (or guard behind `__DEV__`).

**Minor**

- `JSON.parse(fileContent as string)` runs in the `if` condition and again inside `decryptBitwardenJson` downstream — fine, but parsing once into a `const parsed` and reusing it would be a small cleanup.
- `@tetherto/pearpass-lib-vault-core@0.1.1` pulls `@noble/ciphers@^2.2.0` / `@noble/hashes@^2.0.1`, both `engines.node >= 20.19.0`. Worth confirming the bare worklet / CI runtimes meet that floor.

The `--defer @tetherto/swarmconf` additions to all six bundle scripts look consistent and match the stated intent.

ndrkltsk:

nickgr2:
The two items from my prior review haven't been addressed at the new HEAD (`296cbf05`):

1. `src/screens/ImportItems/index.tsx` ~line 290 — `JSON.parse(fileContent as string).encrypted` still runs unconditionally for Bitwarden, but Bitwarden's `accepts` includes `.csv` (line 94). On a CSV upload, `JSON.parse` throws `SyntaxError`, the outer `catch (err)` swallows it, and the user sees "Failed to decrypt file. Please check your password and try again." for a previously working path. Gate on `fileType === 'json'` first.

2. ~Line 306 — `console.log('[bw-import] decrypt failed:', err)` is still there. Please remove (or guard behind `__DEV__`).

The new `--defer @tetherto/swarmconf` additions to the three `*:nightly` bundle scripts in `package.json` are consistent with the non-nightly variants — those look fine.

My `CHANGES_REQUESTED` from `ca7ccdc` still applies; happy to re-review once (1) and (2) are addressed.

ndrkltsk:
@nickgr2 review again please

nickgr2 (approved):
Thanks for the follow-ups — both flagged issues are addressed.

**1. CSV regression** — `fileType === 'json'` gates the parse at `src/screens/ImportItems/index.tsx:287-289`, so CSV uploads no longer hit `JSON.parse`. Verified path.

**2. `console.log` removed** — the catch at line 310 is now `catch { throw new Error(...) }` with no error body logged. Good.

**Verified deps line up at HEAD:**
- `decryptBitwardenJson` and the `decryptViaWorklet` option are exported from `@tetherto/pearpass-lib-data-import` at the new pinned SHA `d6b035dc`
- `pearpassVaultClient.decryptBitwardenExport` is defined at `src/pearpassVaultClient/index.js:830` in vault-core SHA `f974888a`
- `.nvmrc` is 22.12.0, clear of the `>= 20.19.0` floor introduced by `@noble/ciphers@^2.2.0` / `@noble/hashes@^2.0.1`
- `--defer @tetherto/swarmconf` added to the three `:nightly` bundle scripts, matching the non-nightly variants

**Minor (non-blocking)**
- `import { pearpassVaultClient } from '@tetherto/pearpass-lib-vault/src/instances'` is a deep import — the vault package's top-level index only re-exports `setPearpassVaultClient` / `setStoragePath`. Worth adding `pearpassVaultClient` to the public surface in a follow-up so consumers don't have to reach into `src/`.
- `decryptBitwardenJson` still re-parses `fileContent` internally, so we parse twice (once for `.encrypted`, once inside the lib). The lib API takes a string, so this is on the lib side to optimize, not here.

ndrkltsk:
added a warning to inform user that Argon2 import will require some time:
Screenshot 2026-05-15 (10 45 55)

ndrkltsk:
@nickgr2 review again

nickgr2 (approved):
Re-review on top of the prior approval (d9d2347): the refactor extracting `types.ts`/`utils.ts` from `ImportItems/index.tsx` is clean, the new `utils.test.ts` covers the tricky cases (account-restricted Bitwarden vs. password-protected, malformed JSON, missing `kdfType`), and the Argon2 `AlertMessage` is a nice UX touch given Bitwarden's defaults (64 MiB / 3 iters) can genuinely take minutes on mobile. The narrowed catch around `decryptBitwardenJson` / `decryptExportData` / `decryptKeePassKdbx` correctly preserves non-password errors (malformed file, unsupported KDF, worklet/IPC failure) so they aren't misattributed as bad passwords. The added `--defer @tetherto/swarmconf` on the nightly bundle scripts matches the description and aligns them with the non-nightly variants.

minor: `isArgon2BitwardenExport` docstring says "Returns false for any non-Bitwarden export," but the implementation only checks `kdfType === BITWARDEN_KDF_ARGON2`. In practice it's fine (only Bitwarden exports carry `kdfType`, and it's only consulted from `selectedFileInfo?.parsedJson` after `detectIsEncrypted` gated on Bitwarden+passwordProtected), so the behavior matches the docstring for every real input — just slightly under-constrained vs. the comment.

LGTM.

boris91 (dismissed):

torkinos (approved):

PR #416, opened 2026-05-18
Fix Personal-vault fallback on access-revoked.

shavtvalishvili:
## Summary
- When a device is removed from the only remaining shared vault, the receiver wipes local data and creates a fallback "Personal" vault. The lingui macro could not statically resolve `t` through `latest.current` destructuring, so `t\`Personal\`` returned empty and `createVault` rejected with "Invalid vault data: name is required".
- Keep `t` outside the `latest.current` ref so the macro resolves it back to `useLingui()`; read the post-delete vaults list from `latest.current.vaults` for the next-vault lookup.

### DIFF
--- a/src/hooks/useVaultAccessRevoked.jsx
+++ b/src/hooks/useVaultAccessRevoked.jsx
@@ -15,11 +15,6 @@ import { AccessRemovedModalContent } from '../containers/Modal/AccessRemovedModa
 import { useModal } from '../context/ModalContext'
 import { logger } from '../utils/logger'
 
-/**
- * Receive-side handler for "another device removed me from this vault".
- * Wipes local data, recovers to a fresh "Personal" vault when nothing
- * remains, and opens the access-removed modal.
- */
 export const useVaultAccessRevoked = () => {
   const { t } = useLingui()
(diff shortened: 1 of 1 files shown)


torkinos:
@nickgr2 please review

torkinos (approved):

nickgr2 (approved):
Fix targets the right cause: the lingui macro from \`@lingui/react/macro\` requires \`t\` to be a direct identifier from \`useLingui()\` so it can statically transform \`t\`\`Personal\`\`\` at build time — destructuring through \`latest.current\` hides it from the macro and the call produces an empty string. Keeping \`t\` outside the ref resolves it.

Reading \`vaults\` from \`latest.current\` for the post-delete \`next\` lookup is also fine; the every-render \`useEffect\` refreshes the ref, so this picks up the freshest list available to the callback.

minor: the existing tests mock \`@lingui/react/macro\` with \`t: (text) => text\`, so they can't reproduce the original symptom — the bad value only appears after the build-time macro transform runs. Worth keeping in mind that this class of regression needs build-output or E2E verification rather than unit tests.

PR #418, opened 2026-05-18
Refactor import option handling for KeePass and update type naming from KeePassKDBX to KeePassXC

ndrkltsk:


ndrkltsk:

boris91 (approved):

torkinos (approved):

torkinos (approved):

nickgr2 (approved):
The KeePass rename is behavior-preserving as far as I can tell — the previous `KeePassKDBX` branch in the decrypt block was effectively dead code since the KeePassXC import option only accepts `.csv` / `.xml` (the file picker would never feed it a `.kdbx`), and the parse-switch case for `KeePassKDBX` that forced `fileType = 'kdbx'` is no longer needed for the same reason. Renaming the enum to match what the option actually represents (KeePassXC's CSV/XML export, not a KDBX path) is a clear win for readability.

A few notes:

- **String value also changed**: `'keepass-kdbx'` → `'keepass-xc'`. If this enum value is ever persisted (analytics, last-used import option in storage, deep links), the rename is a silent breaking change. Worth a grep for the old string before this lands in a release — `src/screens/ImportItems/index.tsx:84` is the only usage I see in the diff, but I can't see beyond it.
- **Unrelated changes bundled**: the F-Droid docs deletion (`docs/fdroid/build.md`, `docs/fdroid/version-check.md`) and the new `*:dev` bundle scripts in `package.json` are unrelated to the KeePass rename described in the PR title. Three separate PRs would have been easier to bisect and revert if any one of them regresses. minor: future PRs in this area would be easier to review split.
- **F-Droid doc removal rationale unclear**: was the F-Droid build channel itself removed, or just the docs? If the build channel still exists, deleting its docs leaves contributors without a reference. If the channel was removed, that's a notable change that should be called out in the PR description.

Approving — the import refactor itself is clean and the dev bundle scripts are additive.

PR #432, opened 2026-05-25
Refresh paired-devices view after kickDevice

shavtvalishvili:
Changes:
- Explicitly `await refetchVault(vaultId)` after `kickDevice` resolves in `BottomSheetRevokeAccessContent`. The previous flow relied on the autobase `'update'` listener installed by `useVault.fetchVault`, which is replaced any time another `useVault` mounts (single handler per client). Targeted refetch removes the race and keeps the list consistent across siblings that may have stomped the listener.


shavtvalishvili:
Reverting — issue is extension-specific (NM bridge doesn't forward 'update' events). Desktop and mobile have working in-process listeners; refetch here was speculative.

